PR #14, opened 2016-06-06
Refac: Abstracting Board

pafmaf:
This is an attempt to abstact away `BRDFile`. `BoardView` then uses it in a 
more object-oriented way. This comes with some downsides, especially 
considering performance (at least as it is now).

Apart from that most things work as before.

 * Abstract `Board` class to encapsulate/abstract
   - derived `BRDBoard` holds a `BRDFile` instance
   - provides Nets, Parts and Pins in vectors

 * Exported most colors to struct

 * Separated draw procedures for different elements
 
 * Added Net list and Component list
   - double click to select

 * Added annotations layer for possible custom content

WIP:

 * Only using `Board` objects in `BoardView`

----------------------
I tested most of this on MacOS and MinGW by merging piernov's branch.

`Board` is an abstract class where `BRDBoard`, holding a `BRDFile` instance, is an imlementation for BRD files. Other formats should be fairly easy to integrate now.

* *Pin* - any contact on the board (test pads, probably vias, chip pins)
  - a Pin may connect to one Net and belong to one Component
* *Net* - a 'potential' that multiple Pins are connected to (e.g. our friend `pp3v3` or `GND`)
  - holds references to all pins connected to it
* *Component* - any component on the board (connectors, resistors, ICs, etc.)
  - holds references to all pins associated with it

The original implementation is very efficient, since it looped over arrays of structs (directly derived from the source file). But different files have different structures so drawing and _extending those structures, e.g. by annotations_ imo is harder not going full OOP and I hope this implementation using the right data structures can keep up with the original one in near future. It's still quite ok on my machine.

Net list and Component list and other parts probably need some more work.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,6 +3,8 @@ PointerBindsToType: false
 AllowShortFunctionsOnASingleLine: false
 BreakBeforeBraces: Attach
 BreakConstructorInitializersBeforeComma: true
+AlwaysBreakTemplateDeclarations: true
+AlwaysBreakBeforeMultilineStrings: true
 IndentWidth: 4
 TabWidth: 4
 UseTab: ForIndentation
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,16 @@ test
 *.sdf
 .vs
 imgui.ini
+
+# VS cache files
+*.aps
+*.ncb
+*.opendb
+*.opensdf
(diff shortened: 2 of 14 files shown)


MattSturgeon:
Merged as 55e44e23b6587f243f42d462761c55a418da5ab0

PR #23, opened 2016-07-01
Cmake port

MattSturgeon:
Note: This branch includes a squashed version of @pafmaf's PR #14. I could rebase this without if I you really want, but I can't see why we shouldn't merge #14.

I've also taken a lot of inspiration from @piernov's unix port, and I'm hoping to work on integrating that next.

------------------------------

# This transitions from using a Visual Studio project, to using the CMake meta-build system.

This poses a number of advantages.

 - The project meta-data is easier to reason about in simple CMakeLists text files
 - CMake comes with a handy packaging feature. There is now a "PACKAGE" target/project that can be built to generate a Null Soft Installer setup program. (See screenshot and attached installer).
 - Fully compatible with multiple build systems. You can easily generate a Visual Studio build system just by running `cd build` `cmake ..`
 - CMake can even build using `msbuild`, allowing you to use simple commands like `cmake --build . --target package` to build the installer without even opening an IDE.

 - CMake allows us to step towards cross platform support, as it is a cross platform build system.

**I have updated the README to include CMake build instructions** and hints on the different generated build targets.

---------------------------------------------------

Screenshot

Note: to build the installer you will need [NSIS](http://nsis.sourceforge.net/Download) installed and in your `%PATH%`

-------------------------------------------------

Here is the installer built: [OpenBoardView-0.3.0-win32.exe.zip](https://github.com/chloridite/OpenBoardView/files/343069/OpenBoardView-0.3.0-win32.exe.zip)

PR #24, opened 2016-07-05
CMake Take 2

MattSturgeon:
Note, while this is (hopefully) the first step to cross platform in the main branch, **this particular PR is still directx/windows only**. The OpenGL port still needs a little more polish.

#### This builds on #23, taking on suggestions from @piernov.

It also forms the basis of my linux-support work.

 - In the first commit I have replaced the included dependancies (imgui and utf8) with [git submodules] which are automatically fetched by CMake to avoid having to run `git submodule --init `.
 - I have added everything needed to build a WiX (`.msi`) installer using CPack. (requires that the WiX toolset is installed and in PATH)
- Also can build `.zip`s and other packages. The binary is still a single `.exe` file.
 - This enables the foundation for a cross platform build system

#### Structure:

- The build system is controlled by the main `CMakeLists` file, which includes other files (and directories) which keeps aspects of the build system seperate.
  - Including directories (with `add_subdirectory`) will run the `CMakeLists` file in that directory.
- `Project.cmake` deals with metadata like the name and version, as well as most packaging related variables.
- The `asset/` directory adds assets to the build and some resources for installers. A `.desktop` file could go here, for example (once we port to linux).
- The `lib/` directory contains included dependancies `imgui` and `utf8`. It also has `platform_libs.cmake` which deals with adding any system libraries.
   - When we port to GLFW/OpenGL `platform_libs.cmake` could try and find GLFW on the system, `lib/CMakeLists` could then fallback to a git submodule if it wasn't found.
- The `src/` directory contains all the project code and creates the main build target. This could be easily extended later to separate aspects of the build into different subdirectories (e.g. a file-format directory/library).

#### ~~There are still a few points I feel need some work,~~

 - **Testing**. While I've tried to keep the CMake scripts fairly well organized, it has evolved a lot from my initial draft and hence could benefit from lots of real world testing.
    - This is probably best done from the master branch. Honestly the CMake is pretty well organized at this point, works on AppVeyor CI and is fairly well (over) commented
 - [x] **libs vs platform_libs** Currently libs are handled by `lib/CMakeLists` while platform libs are done in the root `CMakeLists`. Decide whether to move platform libs out of the root file and if so where to put them. 
 - ~~**Build options**.~~ - this is really more of an issue in the WIP glfw port.
    - [x] Make generating .zip packages easier
 - [x] **CI**. Add a AppVeyor CI integration file

_P.S. Don't be scared by the 23'000 deletions, it's almost all just replacing libs with submodule equivalents!_

[git submodules]: https://git-scm.com/book/en/v2/Git-Tools-Submodules


chloridite:
Speaking personally, I despise installers, and I always prefer simple .zip distributions or similar when available.  I'd like to keep the availability of a Windows build which is a single "fat" .exe (all the asset dependencies packaged in the .exe as resources, and compiled with the static C runtime).  Yes, this requires platform-specific asset loading code, but the benefit is that if people want to redistribute the program across Windows machines, they can just copy the .exe.

On the Linux side, `pacaur -S openboardview` is my ideal.  Simpler, certainly.

Side note, most machines will have DirectX 9, but if you do have an installer it'd be a good idea to check for it and run the redist installer if necessary.  Another related benefit of an installer is having an OpenGL exe which can be ran as a fallback if d3d9 device creation fails for some reason.

MattSturgeon:
> Speaking personally, I despise installers, and I always prefer simple .zip distributions or similar when available.

No problem, simple exe is available by not building the `PACKAGE` target, or .zip can be automated by setting `CPACK_GENERATOR` to "zip" `cpack -C release -G zip`. I'll try and make this more accessible.



> I'd like to keep the availability of a Windows build which is a single "fat" .exe (all the asset dependencies packaged in the .exe as resources, and compiled with the static C runtime)

Currently (in this part of the port) asset loading remains unchanged. For the GLFW port I've dropped it for now in favour of just using imgui's built in font but that's off topic for this PR. I've made an issue for discussion over on mattsturgeon/openboardview#5.



> Side note, most machines will have DirectX 9, but if you do have an installer it'd be a good idea to check for it and run the redist installer if necessary. Another related benefit of an installer is having an OpenGL exe which can be ran as a fallback if d3d9 device creation fails for some reason.

Good idea! I'll have to look into the best way to impliment that using the [WiX CPack generator](https://cmake.org/cmake/help/v3.0/module/CPackWIX.html)

MattSturgeon:
Feeling like this is ready now. 

Unless anyone has any more feedback (@OpenBoardView/maintainers?) I'll merge this in soon-ish and get back to lin-p4.

README.md
```diff
@@ -1,5 +1,7 @@
 ## Open Board Viewer
 
+AppVeyor Windows build: [![Build status](https://ci.appveyor.com/api/projects/status/pa11fdwisv7x97js/branch/master?svg=true)](https://ci.appveyor.com/project/MattSturgeon/openboardview/branch/master)
```
MattSturgeon: Change this to OpenBoardView account 

```MarkDown
[![Build status](https://ci.appveyor.com/api/projects/status/avgn60acyn0cqyw4/branch/master?svg=true)](https://ci.appveyor.com/project/OpenBoardView/openboardview/branch/master)
```

CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+# Enable solution folders (in IDEs)
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+# Name the solution folder for targets like INSTALL, PACKAGE, ZERO_CHECK
+set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "Meta Targets")
+
+# Which packages should we generate when installing "package" target
+# or running CPack without specifying a generator
+set(CPACK_GENERATOR  WIX ZIP CACHE STRING "List of generators to build packages with")
```
piernov: Isn't it Windows specific? Also sometimes later you redefine it here: https://github.com/OpenBoardView/OpenBoardView/pull/24/files#diff-c6410798361ab4228b2112b31d72c95eR60 .

MattSturgeon: Yes it's Windows specific - it gets put into an if-block in lin-p5, but no harm in setting it here. Good spot on the double definition though!

MattSturgeon:
After some feedback from @piernov I've added `-DImDrawIdx=unsigned int` as a public declaration to imgui (see [`lib/CMakeLists` line 17](https://github.com/OpenBoardView/OpenBoardView/pull/24/files#diff-5b579ba0237255aa2c66982834e4fa5eR17)) and options for easily showing warnings in [`src/CMakeLists`](https://github.com/OpenBoardView/OpenBoardView/pull/24/files#diff-95e351a3805a1dafa85bf20b81d086e6R44).

Setting `ImDrawIdx` to be an `unsigned int` apparently fixes a bug I haven't been able to reproduce. I tried `unsigned long` too, but some code in `imgui.cpp` caused core dumps and I haven't looked into this.

I'm defaulting to setting `SHOW_WARNINGS="normal"` and `WARNINGS_ARE_ERRORS=OFF` for now, I feel this should be changed in a separate PR that also fixes the warnings.

`SHOW_WARNINGS` and `WARNINGS_ARE_ERRORS` currently use [MSVC compile flags](https://msdn.microsoft.com/en-us/library/thxezb7y.aspx). My [WIP GLFW port](https://github.com/MattSturgeon/OpenBoardView/blob/lin-p5/src/CMakeLists.txt#L72) adds gcc/clang flags.

--------------------------------------

Any more feedback from @chloridite or @OpenBoardView/maintainers (or anyone else) would be appreciated, like I said before this is pretty much ready. :smile:

lib/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+### lib/CMakeLists.txt ###
+#
+#
+
+## Platform and system libraries
+include(platform_libs.cmake)
```
piernov: CMake files usually use CamelCase for their names, not a big deal though.

lib/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+	COMMAND git submodule update --init imgui
+	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+add_library(imgui imgui/imgui.cpp imgui/imgui_draw.cpp)
+target_include_directories(imgui PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/imgui")
+target_link_libraries(imgui ${PLATFORM_LIBS})
+target_compile_definitions(imgui PUBLIC "-DImDrawIdx=unsigned int") # short is not enough for us (long causes core dumps)
+set(LIBS ${LIBS} imgui)
+
```
piernov: I may have missed something but I don't think `imgui` is being added to `INCLUDE_DIRS`. Also I prefer exporting variables like `IMGUI_INCLUDE_DIRS` and adding them separaty with `target_include_directories` on the relevent target.

src/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+if (MSVC)
+	target_compile_definitions(${PROJECT_NAME} PRIVATE
+		-D_CRT_SECURE_NO_WARNINGS
+		-D_UNICODE -DUNICODE)
+endif()
+
+# Warnings
+set(SHOW_WARNINGS "normal" CACHE STRING "Compile time warnings to show. Can be \"normal\", \"all\" or \"all_extra\".")
+set_property(CACHE SHOW_WARNINGS PROPERTY STRINGS "normal" "all" "all_extra")
```
piernov: Not sure `set_proprety` is really useful there.

src/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+endif()
+if(WARNINGS_ARE_ERRORS)
+	if(MSVC)
+		set(WARN_FLAGS ${WARN_FLAGS} "-WX")
+	endif()
+endif()
+if(WARN_FLAGS)
+	target_compile_options(${PROJECT_NAME} PRIVATE ${WARN_FLAGS})
+endif()
+
```
piernov: Would have preferred this being defined in the top-level CMake file and I think we could just added/removed `-Wall` and such when we felt like it. But it's ok.

piernov:
Added a few comments (and I still didn't test it) but it looks ok.

src/CMakeLists.txt
MattSturgeon: It let's the GUI show a drop down, or `ccmake` to iterate through valid strings.

src/CMakeLists.txt
MattSturgeon: With MSVC, `-Wall` is equivalent to gcc's `-Wextra`, hence the separation.

The reason I defined it in `src/` is because I didn't want to subject other targets to our warning flags. I could move setting `WARN_FLAGS` to the top level and still apply it to the OpenBoardView target, though.

lib/CMakeLists.txt
MattSturgeon: `imgui` has it's included directories added to the target on line 15. They are `PUBLIC` so targets that add `imgui` will inherit them automatically.

Re vars like `LIBS`, and `INCLUDE_DIRS`; I know they're not perfect - the idea was to keep library related stuff in `lib/` and just add a couple lists to the main target in `src/` without having to edit the main target whenever we add or drop a dependency.

piernov: Yeah sorry, I did miss that.
> without having to edit the main target whenever we add or drop a dependency
In fact I prefer this. But the way you do it is perfectly ok.

lib/CMakeLists.txt
MattSturgeon: Agreed this is out of place, I'll change it to `PlatformLibs.cmake` before merging.

Issue #22, opened 2016-06-22
exe wont launch on two systems


lerocm:
OpenBoardView.exe wont launch on the two systems below. After double clicking the exe, the hourglass icon appears for a brief moment then nothing happens.

Ist system: MSI AP1920 Windtop running Windows 7 Home Premium 64 bit OS, CPU Atom D525 , Integrated graphics Intel GMA 3150, Avira Antivirus running, Zone Alarm, DirectX 11

2nd system: MSI Wind U100 Netbook, WinXP Home Edition SP3, CPU Atom N270, Integrated Graphics Intel GMA 950, Avira Antivirus, Zone Alarm, DirectX 9.0c

All other board viewers run, including the TopTest/Landrex versions and NeoBoardViewer

slayer200230:
1 what ver you trying to run? 2 works on win 10 and linux just fine, avira isnt a problem as i have it.

piernov:
Just tried on an Asus 1201NL (Atom N270) with Windows 10 and it works, so it's not an issue with the CPU. But on this machine the GPU is an NVidia ION. It might be a software issue: broken DirectX 9 or graphics driver for example. You should try to update/reinstall both.

LarryOffley:
I also have a Windows 7 64bit machine that it won't open on. Intel Core 2 Duo With Intel Q35 Video Card. Dual Monitors. It runs on every other machine in the shop but mine. I have reinstalled Direct X done a SFC scan nothing else seems to have a problem. I wonder if some other software is causing a conflict.

slayer200230:
@LarryOffley do you happen to have zone alarm? i think it has to be software. what kinds of antivirus do you have?

LarryOffley:
Avg but that is on machine that it does work on.

I have a lot of open source stuff might be a library conflict I don't know
yet.

Larry

charliebruce:
Try this build out:

[OpenBoardView-ErrorMessages.zip](https://github.com/chloridite/OpenBoardView/files/330650/OpenBoardView-ErrorMessages.zip)

Hopefully it'll show an error message rather than just silently exiting.

LarryOffley:
Having the error is very useful

Error:Failed to create direct 3D device.

Interesting I haven't had any other programs fail. I will try reinstalling
my video drivers anyways.

Larry Offley

On Jun 23, 2016 12:20 PM, "Charlie" <notifications@github.com> wrote:

Try this build out:

[...]

PR #25, opened 2016-07-06
Usability improvements

charliebruce:
Add a bunch of minor usability tweaks:

* Defaults to .brd file type on Linux, OS X and Windows, with an option for "All Files" if your board file has an odd extension (untested on anything except Windows)
* Shows error messages if DirectX fails to initialise
* Adds keyboard controls to move the viewport around (caveat: behaves oddly when rotated at 90 and 270 degrees - will refactor/fix this soon)
* Drag and drop to open file on Windows
* Window title shows filename on Windows

### DIFF
--- a/BoardView.cpp
+++ b/BoardView.cpp
@@ -334,6 +334,39 @@ void BoardView::HandleInput() {
 			target.y *= 0.5f;
 			ChangeZoom(target, 0.5f);
 		}
+
+		//Arrow keys or WASD to move viewport by a percentage of the screen width/height
+		{
+			ImVec2 delta(0.0f, 0.0f);
+			
+			float dist = 0.25f;
+
+			// Ctrl slows down the movement speed
+			if (ImGui::IsKeyDown(17)) {
+				dist *= 0.1f;
+			}
+
+			if (ImGui::IsKeyPressed('W') || ImGui::IsKeyPressed(38))
+				delta.y = dist;
+			if (ImGui::IsKeyPressed('S') || ImGui::IsKeyPressed(40))
+				delta.y = -dist;
+			if (ImGui::IsKeyPressed('A') || ImGui::IsKeyPressed(37))
+				delta.x = dist;
+			if (ImGui::IsKeyPressed('D') || ImGui::IsKeyPressed(39))
+				delta.x = -dist;
+
+			if (delta.x != 0.0f || delta.y != 0.0f) {
+
+				//Account for zoom level by using screen coordinates (TODO easier using m_scale?)
+				ImVec2 dims = ImGui::GetWindowSize();
+				ImVec2 dimc = ScreenToCoord(dims.x, dims.y, 0);
+				ImVec2 c = ScreenToCoord(0.0f, 0.0f, 0);
+				ImVec2 motion((dimc.x-c.x)*delta.x, (dimc.y-c.y)*delta.y);
+				m_dx += motion.x;
+				m_dy += motion.y;
+				m_needsRedraw = true;
+			}
+		}
 		
 		// Flip board:
 		if (ImGui::IsKeyPressed(' ')) {
--- a/BoardView.cpp
+++ b/BoardView.cpp
@@ -28,6 +28,11 @@ BoardView::~BoardView() {
 	free(m_lastFileOpenName);
 }
 
+void BoardView::ShowError(char* msg) {
+	m_lastErrorMsg = msg;
+	m_showError = true;
+}
+
 #pragma region Update Logic
 void BoardView::Update() {
 	bool open_file = false;
@@ -43,7 +48,7 @@ void BoardView::Update() {
 			if (ImGui::MenuItem("Open", "Ctrl+O")) {
 				open_file = true;
 			}
-			if (ImGui::MenuItem("Quit")) {
+			if (ImGui::MenuItem("Quit", "Ctrl+Q")) {
 				m_wantsQuit = true;
 			}
 			ImGui::EndMenu();
@@ -86,9 +91,9 @@ void BoardView::Update() {
 		if (m_showComponentSearch && m_file) {
 			ImGui::OpenPopup("Search for Component");
 		}
-		if (m_lastFileOpenWasInvalid) {
-			ImGui::OpenPopup("Error opening file");
-			m_lastFileOpenWasInvalid = false;
+		if (m_showError) {
+			ImGui::OpenPopup("Error");
+			m_showError = false;
 		}
 		if (ImGui::BeginPopupModal("Search for Net", nullptr, ImGuiWindowFlags_AlwaysAutoResize)) {
 			if (m_showNetfilterSearch) {
@@ -203,9 +208,8 @@ void BoardView::Update() {
 			    "SOFTWARE.");
 			ImGui::EndPopup();
 		}
-		if (ImGui::BeginPopupModal("Error opening file")) {
-			ImGui::Text("There was an error opening the file: %s", m_lastFileOpenName);
-			// TODO: error details? -- would need the loader to say what's wrong.
+		if (ImGui::BeginPopupModal("Error")) {
+			ImGui::Text("There was an error: %s", m_lastErrorMsg);
 			if (ImGui::Button("OK")) {
 				ImGui::CloseCurrentPopup();
 			}
@@ -937,7 +941,7 @@ void BoardView::OpenFile(char * filename)
 			m_wantsTitleChange = true;
 		}
 		else {
-			m_lastFileOpenWasInvalid = true;
+			ShowError("Cannot parse the file.");
 			delete file;
 		}
 		free(buffer);
--- a/BoardView.cpp
+++ b/BoardView.cpp
@@ -334,38 +334,58 @@ void BoardView::HandleInput() {
 	}
 	if (!io.WantCaptureKeyboard) {
 
-		// (I/+) and (O/-) keys as alternate zoom. Will zoom on the centre of the viewport, not on the mouse.
-		if (ImGui::IsKeyPressed('I') || ImGui::IsKeyPressed(187)) {
-			ImVec2 target = ImGui::GetWindowSize();
-			target.x *= 0.5f;
-			target.y *= 0.5f;
-			ChangeZoom(target, 2.0f);
+		//Ctrl-Q as alternative quit shortcut (Alt-F4 is already handled by Windows)
+		if (ImGui::IsKeyPressed('Q') && ImGui::IsKeyDown(17)) {
+			m_wantsQuit = true;
 		}
-		if (ImGui::IsKeyPressed('O') || ImGui::IsKeyPressed(189)) {
+
+		// (I/+/PgUp) and (O/-/PgDn) keys as alternate zoom. Will zoom on the centre of the viewport, not on the mouse.
+		
+		//TODO numpad +/- scancodes
+		if ( ImGui::IsKeyPressed('I')
+			|| ImGui::IsKeyPressed(187)
+			|| ImGui::IsKeyPressed(ImGui::GetKeyIndex(ImGuiKey_PageUp))
+			)
+		{
 			ImVec2 target = ImGui::GetWindowSize();
 			target.x *= 0.5f;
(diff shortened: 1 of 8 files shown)


inflex:
Already have the keyboard pan controls sorted out to deal with the 90/270/flip problem.

I'm just waiting for a stable core base before I start rolling all my fixes back in to it.

MattSturgeon:
@charliebruce Thanks for taking the time to refactor your changes, and adding file filters to the other `show_file_picker` implementations! I'll get around to looking at this within a few hours.

@inflex were your changes based on @charliebruce's original PRs? If so we should be able to get this in and then add your patches on top with the GLFW port.

It'd be great if someone with OSX access could test 61ec392fe885588684affb09e991adc04ff1ca50 on a branch that builds for mac. (@piernov ?)

piernov:
61ec392fe885588684affb09e991adc04ff1ca50 seems to work on OS X after cherry-picking it into my repo.

Since we support other file format it would be nice to put the file extensions in a vector (or even a map for giving a reference to the function handling it) and loop through it. No need to hardcode them in 4 different places.

MattSturgeon:
> 61ec392 seems to work on OS X after cherry-picking it into my repo.

Good stuff

> Since we support other file format it would be nice to put the file
extensions in a vector

I've seen it done like so:

(Will edit in from pc)

MattSturgeon:
@charliebruce 

#### Feedback on the keyboard shortcuts:

 - For me personally the "default" jump distance is a bit too far, I find myself uncomfortably holding `ctrl` all the time! I's suggest you could decrease the default movement to be closer to slow-mode, and add a fast-mode toggled by `shift`.
 - The zoom in/out has no speed adjustment. Keyboard/mouse zoom speed should work the same as panning around.
   - `I`/`O` will probably need changing as `Ctrl+O` maps to open.
   - Numpad `+`/`-` should work the same as normal `+`/`-`
 - Would be good to add `Ctrl+Q` for quitting

Otherwise nice work!

#### Other bits

 - Drag 'n' Drop: Works well!
    - I'd prefer if the "multiple files" error message was consistent with other dialogues like the "Error opening file" one (see below)
 - Could you abstract `MessageBox` into a `show_error_msg` wrapper in the platform files as this looks windows specific.
   - Also would be good to move the other errors (I can think of "error loading file") to use this new function.

 - I was unable to test the d3d error messages, but I assume it's the same as the one reported working in #22. I don't know yet how much of this will be useful once we've dumped dx9 for GL, but it's affecting users now so

> keyboard controls to move the viewport around (caveat: behaves oddly when rotated at 90 and 270 degrees - will refactor/fix this soon)

It would be great if you got this sorted, I believe @inflex had done some work on this that could help (albeit based off an early linux port)

-----------------------

I quite like this on the whole, the `BoardView::ChangeZoom` and `BoardView::OpenFile` abstractions are nice too.

Final notes for now, it'd be good if you rebased out some of the extra commits from the history (like the merge commits). Also I haven't checked, but have you run [`clang-format.py`](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/) on your changes yet?

-------------------------------
@piernov 
> Since we support other file formats it would be nice to put the file extensions in an std::vector (or even an std::map for giving a reference to the method handling it) and loop through it. No need to hardcode them in 4 different places.

Misread this earlier. Yes, it would be good to move to a unified filter function. I'm not sure this is needed for the initial implementation though.

pafmaf:
> `[op setAllowedFileTypes:@[@"brd", @"other", @"file", @"types"]];`

As I understood it

> for giving a _reference_ to the _method_ handling it and loop through it

means having something like a factory pattern with `Board`s, i.e. different file types have to register somewhere and the factory _method_ instantiates the right thing. `Board::getRegisteredFileTypes()`

MattSturgeon:
@pafmaf @piernov 

> As I understood it
> 
>     for giving a reference to the method handling it and loop through it
> 
> means having something like a factory pattern with Boards, i.e. different file types have to register somewhere and the factory method instantiates the right thing. Board::getRegisteredFileTypes()

Yes I missed that he was talking about a map/filter functions - doh! (see the note at the end of my last [comment](#issuecomment-230737865) - added it just as you posted!)

charliebruce:
Have made some changes:
* Have added Ctrl-Q as a quit shortcut.
* Refactored error-handling.
* Added numpad controls, also PgUp/PgDn zoom in/out.
* Have halved default movement speed, with Ctrl working as before (decreasing movement speed by factor of 10). Feels easier to use now. I personally prefer Ctrl-to-slow-down rather than Shift-to-speed-up, is there a standard/expected behaviour in the industry or is this up to preference?

I haven't fully addressed all points though:
* Haven't added speed control on keyboard zooming - that level of precision when zooming doesn't feel necessary to me, and allows the idiom I/O for In/Out to be kept, without breaking the Ctrl-O shortcut. As before, if there's an expected behaviour/use case for it can easily add though.
* Haven't tested numpad yet - my primary machine is a laptop so I don't have one to test with unless I dig up an external USB one.
* Have kept the Direct3D errors in MessageBoxes because they occur before ImGui has loaded so can't rely on it to show errors, and will necessarily be Windows-specific anyway.
* Probably not worth refactoring file type filtering for now, can sort that once we have a larger set of supported file types/importers.

Will do over weekend:
* 90/270 rotation fix.
* Don't have Python set up on this machine, haven't run clang-format.py yet.
* Rebase/squash to tidy everything up.

BoardView.cpp
```diff
@@ -334,38 +334,58 @@ void BoardView::HandleInput() {
 	}
 	if (!io.WantCaptureKeyboard) {
 
-		// (I/+) and (O/-) keys as alternate zoom. Will zoom on the centre of the viewport, not on the mouse.
-		if (ImGui::IsKeyPressed('I') || ImGui::IsKeyPressed(187)) {
+		//Ctrl-Q as alternative quit shortcut (Alt-F4 is already handled by Windows)
+		if (ImGui::IsKeyPressed('Q') && ImGui::IsKeyDown(17)) {
```
MattSturgeon: Could you move this up to line 41 so it is together with `Ctrl+O`?

BoardView.cpp
```diff
@@ -334,38 +334,58 @@ void BoardView::HandleInput() {
...
+			m_wantsQuit = true;
+		}
+
+		// (I/+/PgUp) and (O/-/PgDn) keys as alternate zoom. Will zoom on the centre of the viewport, not on the mouse.
+		
+		//TODO numpad +/- scancodes
+		if ( ImGui::IsKeyPressed('I')
+			|| ImGui::IsKeyPressed(187)
```
MattSturgeon: If possible, could you replace all hard coded numbers with `GetKeyIndex`?  That should aid readability (and porting to GLFW later)

BoardView.cpp
```diff
@@ -28,6 +28,11 @@ BoardView::~BoardView() {
 	free(m_lastFileOpenName);
 }
 
+void BoardView::ShowError(char* msg) {
+	m_lastErrorMsg = msg;
```
MattSturgeon: Great, this refrac should make it easier to drop `ImGui::OpenPopup` for errors and just use native dialogs for all errors.

Would mean wrapping `MessageBox`, and Gtk + Cocoa equivalents in a cross-platform `show_error` function though, so happy to leave it for now.

platform_unix.cpp
```diff
@@ -51,6 +51,11 @@ char *show_file_picker() {
 					"_Open", GTK_RESPONSE_ACCEPT,
 					NULL );
 
+	// Filter file types
+	GtkFileFilter *filter = gtk_file_filter_new();
+	gtk_file_filter_add_pattern(filter, "*.brd");
+	gtk_file_chooser_add_filter(dialog, filter);
```
MattSturgeon: This needs to be `gtk_file_chooser_add_filter(GTK_FILE_CHOOSER(dialog), filter);` - turns out the dialog needs casting to the `GtkFileChooser` interface to use `gtk_file_chooser_add_filter` on it.

MattSturgeon: For feature parity with windows, you should also use `gtk_file_filter_set_name(filter, "BRD Files");` and consider adding a second GtkFileFilter with pattern `"*"` and name `"All Files"`

BoardView.cpp
```diff
@@ -320,22 +326,87 @@ void BoardView::HandleInput() {
...
+			float factor = powf(2.0f, mwheel);
+			ChangeZoom(target, factor);
 		}
 	}
 	if (!io.WantCaptureKeyboard) {
+
+		//Ctrl-Q as alternative quit shortcut (Alt-F4 is already handled by Windows)
+		if (ImGui::IsKeyPressed('Q') && ImGui::IsKeyDown(17)) {
+			m_wantsQuit = true;
+		}
```
MattSturgeon: Could you have `Ctrl+O` and `Ctrl+Q` together? Would probably mean moving this to the top of  `Update` method.

platform_osx.mm
```diff
@@ -7,6 +7,10 @@
 	std::string filename;
 	NSOpenPanel *op = [NSOpenPanel openPanel];
 
+	// Filter file types
+	[op setCanChooseFiles:YES];
+	[op setAllowedFileTypes:@[@"brd"]];
```
MattSturgeon: May be worth adding `[op setAllowsMultipleSelection:FALSE];`

MattSturgeon:
Hi @charliebruce, appologies if you've been put off by some of the feedback (or the volume of it). I'd still like to get this in, are you still planning to work on this?

If not I am happy to make the final adjustments needed to get these changes merged.

charliebruce:
Hi @MattSturgeon - no problems with the feedback, have just been quite busy with other projects over the last couple of weeks. I can take another look at all this stuff on Friday/Saturday evening, unless you have any free time before then and want to get it done sooner - in that case, feel free to build on my changes or cherry-pick anything that is useful to you.

MattSturgeon:
@charliebruce I've split out the keyboard stuff from the rest in [charlie-part1] and [charlie-part2]. I've invited you to a new "Charliebruce PR" team to give you push access to the new branches.

Feel free to work off of those branches and replace this PR.

I've not tested these yet, so hopefully I haven't messed up anything with my rebasing. (I'll edit this line once I've tested the new branches)

[charlie-part1]: https://github.com/OpenBoardView/OpenBoardView/commits/charlie-part1
[charlie-part2]: https://github.com/OpenBoardView/OpenBoardView/commits/charlie-part2

MattSturgeon:
@charliebruce I've merged part1 and added some changes to part2.

### [Changes to part2](https://github.com/OpenBoardView/OpenBoardView/compare/b42316e...charlie-part2)

- Moved `Ctrl+O` from `Update()` into `HandleInput()`
  - The `open_file` variable is now the property `m_open_file`
- Added a generic speed modifier
  - `Ctrl` will divide speed by 5 (`speed = 0.2`)
  - `Shift` will multiply speed by 5 (`speed = 5`)
  - Tweaked default `distance`, `zoom` and `mwheel` and applied `speed` to each of them.
- Minor changes
  - Fixed brackets around checking `GetKeyIndex` (fixes page up/down)
  - Use `io.KeyCtrl` instead of `IsKeyDown(17)`
  - moved `powof` into `ChangeZoom()`: positive `factor` will zoom in, negative out. This makes it easier to apply `speed` modifier.

See the [diff](https://github.com/OpenBoardView/OpenBoardView/compare/b42316e...charlie-part2) for full changes

### [Merged into master](https://github.com/OpenBoardView/OpenBoardView/commits/master)

- Refactor error handling
- Display an error when DirectX can't be initialised
  - Rather than silently exiting, we should inform the user if DirectX can't
     be initialised.
- Add drag and drop
- Display filename in window title
- Add .brd filetype filter to the Open File dialog
  - On Windows and Gtk (linux) we add a drop down to choose between any file
     or just .brd files. On OSX we tell the dialog that we only want .brd
     files.

MattSturgeon:
Merged the last part as 4d245d94be0807c9402a207ad1cf8e7d57f5fd95. In addition to the changes noted above, I also changed a few comments and reordered the contents of `HandleInput()` to group all shortcuts together and all controls together.

PR #29, opened 2016-07-12
Dds replacement

inflex:
Removes the need for DDS/textures.  This is the simple starting point for all the rest of the vector draw code that has been generated in a separate fork.


MattSturgeon:
Working well on linux, when merged into lin-p5.

screenshot from 2016-07-12 09-10-22

Can you remove the following redundant code:

- dds file from `asset/` (use `git rm asset/*.dds`)
- line `12` from the  `.rc` file
-  Lines from  `src/imgui_impl_dx9.cpp`:
  - `ImGui_ImplDX9_CreateAssetTexture()` lines `329`-`362`
  -  `ImGui_ImplDX9_CreateCircleTexture()` lines `363`-`369` 
  -  lines `375`-`365`
- Lines `4` and `6` from `src/resource.h`

src/imgui_impl_dx9.cpp
```diff
@@ -326,54 +326,10 @@ struct DDSHeader {
...
-	result &= ImGui_ImplDX9_CreateAssetTexture(1, ASSET_EMPTY_CIRCLE);
-	return result;
-}
 
 bool ImGui_ImplDX9_CreateDeviceObjects() {
 	if (!g_pd3dDevice)
 		return false;
-	if (!ImGui_ImplDX9_CreateFontsTexture())
-		return false;
-	if (!ImGui_ImplDX9_CreateCircleTexture())
```
MattSturgeon: The first two lines should stay

```
if (!ImGui_ImplDX9_CreateFontsTexture())
    return false;
```

piernov:
You should also delete the following lines in BoardView.cpp:

```
414:	ImTextureID filled_circle_tex = TextureIDs[0];
415:	ImTextureID empty_circle_tex = TextureIDs[1];
436:	auto pin_texture = empty_circle_tex;
464:	// pin_texture = filled_circle_tex; // TODO
```

MattSturgeon:
> And TextureIDs[] might not be needed anymore.

Tested removing `TextureIDs[]`, etc in [this patch](http://pastebin.com/65BC5Wn5) and seems to work fine.

piernov:
`DDSHeader` and `DDSPixelFormat` structs can be removed too, along with `typedef unsigned long u32;`

chloridite:
A note from my original development: the poly count from AddCircle made rendering slow as hell, and having an image with hand-drawn mipmaps allowed me the control I desired over the circles' line thickness.

inflex:
@chloridite interesting, performance here is very slick, even when loading up macbookpro boards (which are obviously very populated).  Setting the segment count between 8~32 probably helps.  I've been running only on generic built in Atom and G3420 default Intel graphics at 1920x1200 / 1600x900.

pafmaf:
> made rendering slow as hell

---

> performance here is very slick

The build is definitely slower over here on several machines. So I suggest using some hybrid solution for this?

MattSturgeon:
@pafmaf is it noticeably slower or just measurably?

e.g. is there any input lag or low framerate that wasn't there before?

pafmaf:
@MattSturgeon

It's _very_ noticeable in debug builds on Win, release builds are not as bad, but I can still make out the difference between the older and newer builds on my (fairly old) desktop. It's got a reasonable GPU for its age and we're only drawing 2D stuff.

I didn't compare it to older builds on my (again, fairly old Core-i3) linux laptop, but it is defintely lagging more that it should, but I'd check that sometime.

piernov:
Maybe rendering to a texture once and then using this texture to actually draw the pads would help? It combines both approaches. But probably hard to do with ImGui…
On a Celeron N2830, 60FPS but drops to ~20FPS when moving the board. (and on a A20 android tablet, drops to ~4FPS but that's not this only commit's fault) With the DDS texture, it runs at 60FPS steady on the Celeron.

pafmaf:
> But probably hard to do with ImGui…

Why? I'm not goot at graphics stuff, but it sounds very doable: wouldn't a method like `drawCircle(float radius, bool filled)` drawing to a buffer / texture, then potentially redrawing that texture on zoom or mipmap suffice?

Also my thoughts on:
> noticeably slower or just measurably

I think that 'measurable' will inevitably mean noticeable for some, e.g. on a `A20 android tablet` . *The* cool thing about OBV (compared to other software) is that it's got HW support.

piernov:
> wouldn't a method like drawCircle(float radius, bool filled) drawing to a buffer / texture,

The problem is that it's not actually drawing anything. It's only adding the command to the list of things to render when calling `ImGui::Render()` (which will call the rendering functions defined in the `imgui_impl_*` renderer). Well, maybe if we bind our GL framebuffer (dunno how Dx9 works), call drawCircle (and nothing before) then Render, it'll render to our buffer.

inflex:
Interesting results all around.  I wonder what's dragging it down for some (I can understand on the Android).  I can't get mine to shift off 60fps more than perhaps at worst 50fps.

Maybe because I'm testing on my inflex-ui-features fork which still is using the SDL backend?

Realistically any modern system should be able to render a few hundred (even thousand) circles quite quickly even on generic GFX hardware.  

The rest of my vector draw features don't depend explicitly on the circle drawing (other than the variable radius feature), so you can flip back to the DDS, it just seems less clean when you start zooming in more.

pafmaf:
> I wonder what's dragging it down for some .. Realistically any modern system should be able to render a few hundred (even thousand) circles

is basically answered by:

> The problem is that it's not actually drawing anything. It's only adding the command to the list of things to render when calling ImGui::Render()

Here are some results from **release** builds while constantly dragging the board around. Debug builds have too detailed profiling output presets, so those would show _vector<>_ iteration from my refac PR, basically taking as long as AddCircle. (This could be a problem on platforms where vectors aren't optimized as good.)

After the DDS PR, CPU usage has jumped up to around 25% whereas before it was around 5% and nearly idleing.

## Before DDS
* CPU around 5%
  * DrawBoard 56%
    - DrawPins 32%
    - DrawOutline 23%
       - **ImDrawList::AddPolyLine** 13%

before

## After DDS
* CPU around 25%
  * DrawBoard 27%
    - DrawPins 25%
      - AddCircle 20%
         - **ImDrawList::AddPolyLine** 8.74%
    - DrawOutline 2.52%

after


>  it just seems less clean when you start zooming in

We could fix that,- not implying we should go back to DDS.

inflex:
For the sake of amusement - change the AddCircleFilled to just AddCircle (maybe a slightly different colour) and see how much of an impact (if any) it makes.

inflex:
I just dug down in to ImGui and looking at the code for circle, and more specifically circle-filled seems a bit less than optimal (for perfect circles), more so in that the filled circle invokes antialiasing too.

I'll look to see if I can implement a replacement pair of calls that are specifically optimal for perfect circles.

MattSturgeon:
Interesting stuff. Hopefully more optimal circle functions will help, will be interesting to see how much.

I'm also interested in the difference between this patch on the dx9, glfw3 and sdl2; from a quick glance at the stats above, "external code" seemes to be a much bigger factor when using imgui's circle functions.

I don't have any lower end machines with windows on them, so I'll struggle to test this.

As a side note, could you move discussion to a new issue specific to the performance hit? Thanks!

inflex:
Add the following in main_opengl (or dx9?) just before "BoardView app{};" ImGuiStyle &style = ImGui::GetStyle();
<inflex>  style.AntiAliasedShapes = false;

inflex:
Add the following in main_opengl (or dx9?) just before "BoardView app{};" 

ImGuiStyle &style = ImGui::GetStyle();
style.AntiAliasedShapes = false;

That turns off AA for shapes (ie, circles) which should greatly improve matters given the number of circles being generated.  I didn't turn it off for segments because they're fairly minimal in count and they do look notably better when AA'd.

This is something that could really be added as a CLI or configuration parameter, ie, --disable-aa

ImGui's Orcornut is speaking of improving some things in the future too.

pafmaf:
> difference between this patch on the dx9, glfw3 and sdl2; from a quick glance at the stats above, "external code" seemes to be a much bigger factor when using imgui's circle functions

@MattSturgeon You might want to check [AddPolyLine](https://github.com/ocornut/imgui/blob/master/imgui_draw.cpp#L412-L591) - there's not much external, it's just labeled `[External Code]` because of the way I profile the _release build_.
In this case I think it's just `ImInvLength`, so again, this is just adding to the list, not really drawing- probably no big difference between dx / sdl.

> I don't have any lower end machines

See:
> Celeron N2830, 60FPS but drops to ~20FPS

Maybe 'lower end' but this is a modern laptop CPU, and dropping to 20FPS(!) due to this commit is not a performance only issue, more a usability thing really.

PR #32, opened 2016-07-14
cmake: Add uninstall target

MattSturgeon:
Add a target that removes all files listed in `install_manifest.txt` (i.e. all files installed by the `install` target).

This is useful if you want to build the `install` target instead of using a package or running the binary from the build directory.

This is more intended for the unix stuff, where people are more likely to be silly and use `make install`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,6 @@ trim_trailing_whitespace = true
 
 
 # Code files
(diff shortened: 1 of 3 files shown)


MattSturgeon:
I couldn't get this to work having tested it just now. Was getting errors from make saying that the target recipe had failed.

May investigate at some point.

PR #34, opened 2016-07-18
Fix an issue where we lose focus after menu items are used.

### LINKED ISSUE #30: Cannot zoom after closing a dialog
After closing a dialog like "About" or "Net" OBV will not respond to the scroll wheel.

If you then re-position the board using the mouse, the scroll wheel will work again.

inflex:
Restructured the window layout so that there's no overlapping between
the main menu bar, board drawing area, and status bar.

Check for hover instead of focus in BoardView::HandleInput(). This
also means that zooming will work when OpenBoardView is not the
focused application.

Fixes #30

### DIFF
--- a/src/BoardView.cpp
+++ b/src/BoardView.cpp
@@ -31,14 +31,17 @@ BoardView::~BoardView() {
 #pragma region Update Logic
 void BoardView::Update() {
 	bool open_file = false;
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Works well on my windows box. Will merge this if no one has any objections.

PR #35, opened 2016-07-18
Adjust clang-format

MattSturgeon:
I've made some **adjustments to the clang-format** file. See the diff for the effect of these changes. Hopefully this should make a few people less frustrated with the current formatting.
- Tidy up: Added comments and grouped options in .clang-format
- Allow short statements to occupy a single line
  - e.g. `if (true) return false;` is now allowed.
- Set case lables to be indented within the switch statement.
- Set grouped assignment and operators to be aligned.
- Align line escapes as far left as possible (rather than at column
  100)
- Bracheted arguments will each have their own line if they cannot
  fit together on one line. They will also align after the opening
  bracket.

---

If there is anything you would like changing I recomend reading through the [cmake-format docs](http://clang.llvm.org/docs/ClangFormatStyleOptions.html) to get an idea of the options and then use `clang-format -i -style=file src/*.{h,cpp}` and `git diff` to see the changes. You can use `git reset --hard` to remove all changes since the last commit.

There are a number of useful clang-format tools; the main binary shown above, a git extension (`git clang-format -h`) useful for only formatting uncommitted changes , vim extension (`Ctrl+K`) and probably a plugin for whatever editor you use. An editorconfig plugin is also useful to have your editor show tabs as intended.

You can find the tools on the [LLVM svn repo](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/); the main binary is usually also included in distro packages.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,11 +1,35 @@
+# .clang-format (YAML)
+
 BasedOnStyle: LLVM
-PointerBindsToType: false
-AllowShortFunctionsOnASingleLine: false
+
+# Indentation
+IndentWidth: 4
+TabWidth: 4
+UseTab: ForIndentation
+
+# Used for line wrapping and defining a "short" statement
+ColumnLimit: 100
+
+# Allow short statements on one line
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: true
+
+# Sensible alignments
+AlignConsecutiveAssignments: false
+AlignOperands: true
+AlignEscapedNewlinesLeft: true
+AlignAfterOpenBracket: Align
+BinPackArguments:  false # Break after each arg if they won't fit on one line
+BinPackParameters: false
+
+# Line breaks
 BreakBeforeBraces: Attach
 BreakConstructorInitializersBeforeComma: true
 AlwaysBreakTemplateDeclarations: true
 AlwaysBreakBeforeMultilineStrings: true
-IndentWidth: 4
-TabWidth: 4
-UseTab: ForIndentation
-ColumnLimit: 100
+
+# Misc
+PointerBindsToType: false
+IndentCaseLabels: true
--- a/src/BRDBoard.cpp
+++ b/src/BRDBoard.cpp
@@ -57,8 +57,7 @@ BRDBoard::BRDBoard(const BRDFile *const boardFile)
 			net->name = string(brd_nail.net);
 
 			// avoid having multiple UNCONNECTED<XXX> references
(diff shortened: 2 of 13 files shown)


inflex:
Time to start some format wars ;)

With the new format, I get lines formatted as follows;
 `ImVec2(pos.x - 1, pos.y - 1), ImVec2(pos.x + 1, pos.y + 1), color);`

Where's I have a personal preference for binding the operator to the variable, but that's a throwback from old Math of school, ie, 

`ImVec2(pos.x  -1, pos.y -1), ImVec2(pos.x +1, pos.y +1), color);`
 
Secondly,  I am finding the operator alignment within a block of code perhaps a bit too much, so maybe go back to non-aligned again, it flows better (to my mind/eyes) without, or find a way to pull back the limit, as there's some lines where there's over 10 spaces gapping.

```
		if ((!jump) && (fp.x == pb.x) && (fp.y == pb.y)) {
			if (i < outline.size() - 2) {
				fp   = *outline[i + 2];
				jump = 1;
				i++;
			}
```

Thirdly (lastly?),  what was the 100 column limit imposed by again?  I am more partial to 120~132, again though, greybeard'ing and longing for my old VT terms in high res mode :p

MattSturgeon:
> I have a personal preference for binding the operator to the variable

Sometimes I agree with this, but usually I like to have a space around infix operators. Either way, I think clang-format is guilty of far more significant crimes and this is probably too subjective to bother changing anyway. (I'm fairly sure this isn't new behaviour in this PR either)

> I am finding the operator alignment within a block of code perhaps a bit too much, [...] there's some lines where there's over 10 spaces gapping.

I'm not sure if there's a way to limit how far over to align. I agree this is an issue, you're probably right that it is better to sacrifice assignment alignment if we cannot limit the indent.



> What was the 100 column limit imposed by again? I am more partial to 120~132

This was an arbitrary limit set by @chloridite, it's probably just the width of his editor.

Personally I like to keep most stuff down to around 80 columns to fit in smaller windows but I'm happy to stick with 100 as a compromise.

At least clang-format is fairly reasonable in how it breaks up long lines; one of the changes here is to make function arguments take up one line each if the original line was too long.

MattSturgeon:
Rebased on master and set `AlignConsecutiveAssignments` to `false` to prevent the ridiculous alignments.

piernov:
How does this compare to the one we just merged from @inflex ? Are we ok with the current formatting?

inflex:
@piernov will look at it in a few hours and see.  Thanks again for the work done, what a merge!

MattSturgeon:
This was an attempt to write a .clang-format file that we could all roughly agree on. Feel free to iterate on it if you like

MattSturgeon:
Closing since no longer relevant.

PR #51, opened 2016-10-27
Feat/search suggestions

piernov:
Please test before merging. Things may be broken.


inflex:
My testing so far has show it to be working "as expected"; not to say there aren't any hiccups but I'm pretty happy with the work done.  The move to templates for the search system leaves me a little out of my depth for now but let's see how this goes.  

Thanks for the improved search suggestions system :+1:

PR #64, opened 2016-12-20
add bv2raw.c, mod to create valid output

corbolais:
thx


corbolais:
@pd: would you want to merge it this way?

PR #71, opened 2017-04-03
Added single-parameter file load mode so that associated filenames/types

inflex:
can be loaded in Windows by the double-click method (without this fix it won't work by default because OBV expects -i <file>

### DIFF
--- a/src/openboardview/main_opengl.cpp
+++ b/src/openboardview/main_opengl.cpp
@@ -90,6 +90,21 @@ char help[] =
 int parse_parameters(int argc, char **argv, struct globals *g) {
 	int param;
 
+
+	/*
+	 * When we're using file-associations, the OS usually just
+	 * passes the filename to be loaded as the single initial
+	 * parameter, so in this special case situation we see if the
+	 * single param is a valid file, and try load it.
+	 */
+	if (argc == 2) {
+		if( access( argv[1], F_OK ) != -1 ) {
+			g->input_file = argv[1];
+			return 0;
+		}
+	}
+
+
 	/**
 	 * Decode the input parameters.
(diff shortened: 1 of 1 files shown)


piernov:
Would be better to add the association information to the Windows registry. (As well as adding a desktop file and mime-type for XDG-compatible Unixes.)

Still nice to have this for easy command line call.

inflex:
It would be better, that I agree; though since we don't have a proper installer I'm suspecting it does make it a bit tricky (unless there's something we can add in OBV much like other Windows apps constantly ask "Do you want us to associate files with this program?".  Ideas?

For now the 1-parameter change is working that I've been able to test (W7, 32bit)

piernov:
For example adding a menu entry which opens a dialog box with checkboxes for each file types? For Windows only though, wouldn't be working on Linux. There might be a Windows API for this purpose.

inflex:
Yes, something within the OBV menu setup was what I was leaning to as an option.  Would such APIs be available/accessible in the cross-compiled build? (I suppose they should be?).  Not so worried about the other OS's atm.

piernov:
For Windows it seems you need to manually edit the registry using the Reg API. Should be available in the MinGW headers too.

About Linux (with XDG standard) it's a matter of defining MIME types for the various formats and adding a .desktop file for OpenBoardView. File association is done at the user's discretion using his Desktop Environment utilities.

No idea about OS X though, may have to do with the application bundle itself.

inflex:
Whoops, looks like I jumped the gun.  I merged the branch in to my master (inflex) and then did a PR to OBV master;  I'm fine either way, just let me know which is the "best" way.

Issue #79, opened 2017-07-20
OBV_BUILD set to "R" from release tarball


srakitnican:
```diff
From c4ac204b333eefca178c345c1e6015c90031cf4f Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Samuel=20Rakitni=C4=8Dan?= <samuel.rakitnican@gmail.com>
Date: Thu, 20 Jul 2017 12:30:52 +0200
Subject: [PATCH] CMake: Allow OBV_BUILD as a build option and use
 PROJECT_VERSION if there is no git

If there is no .git directory for example from a release tarball, OBV_BUILD is
set to "R". This patch should determine if git tree is present and if not use
PROJECT_VERSION as OBV_BUILD.

Also allow to set custom OBV_BUILD from command line or configuration.

---
 CMakeLists.txt | 50 ++++++++++++++++++++++++++++----------------------
 1 file changed, 28 insertions(+), 22 deletions(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8f7375a..ea820f0 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,28 +11,34 @@ if(DEFINED ENV{TERM})
 endif()
 
 # Build info
-execute_process(
-	COMMAND git rev-list HEAD --count
-	OUTPUT_VARIABLE GIT_REVISION
-	OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-execute_process(
-	COMMAND git name-rev --name-only HEAD
-	OUTPUT_VARIABLE GIT_BRANCH
-	OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-execute_process(
-	COMMAND git config branch.${GIT_BRANCH}.remote
-	OUTPUT_VARIABLE GIT_REMOTE
-	OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-execute_process(
-	COMMAND git config remote.${GIT_REMOTE}.url
-	OUTPUT_VARIABLE GIT_URL
-	OUTPUT_STRIP_TRAILING_WHITESPACE
-)
-string(REGEX REPLACE ".*/(.+)/.+$" "\\1" GIT_REPO "${GIT_URL}") # Extract the user from a Github repo URL
-set(OBV_BUILD "R${GIT_REVISION} ${GIT_REPO}/${GIT_BRANCH}")
+if (NOT DEFINED OBV_BUILD)
[...]

PR #80, opened 2017-07-20
CMake: OBV_BUILD as a build option and use PROJECT_VERSION

srakitnican:
if there is no git

If there is no .git directory for example from a release tarball, OBV_BUILD is
set to "R". This patch should determine if git tree is present and if not use
PROJECT_VERSION as OBV_BUILD.

Also allow to set custom OBV_BUILD from command line or configuration.

Issue #79

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,28 +13,34 @@ if(DEFINED ENV{TERM})
 endif()
 
 # Build info
-execute_process(
-	COMMAND git rev-list HEAD --count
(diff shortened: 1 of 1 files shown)


srakitnican:
This will not work, it has old changes.

piernov:
You don't need to open a new PR on Github. It automatically updates the PR with the changes from the branch you specified. Knowing that, you should work on separate branches in your repo (not `master`) so you can open 2 PR and work on your own for example.

PR #81, opened 2017-07-20
CMake: OBV_BUILD as a build option and use PROJECT_VERSION if there is no git

srakitnican:
If there is no .git directory for example from a release tarball, OBV_BUILD is
set to "R". This patch should determine if git tree is present and if not use
PROJECT_VERSION as OBV_BUILD.

Also allow to set custom OBV_BUILD from command line or configuration.

Issue #79


piernov:
Thanks a lot for the contribution! Closes #79

Issue #33, opened 2016-07-17
Use file association and default programs


MattSturgeon:
Closely related to #27, we should set up OpenBoardView to be associated with `.brd` files and ask to be the default application.

On freedesktop.org systems this involves registering the `.brd` `application/x-brd-board` mimetype (as we will do for icon association anyway) and then list that mimetype in the `.desktop` file. Once we run `update-mime-database` and `update-desktop-database` The system will know to use `OpenBoardView.desktop` for `.brd` files.

On windows the relavent docs are [here](https://msdn.microsoft.com/en-us/library/windows/desktop/cc144156(v=vs.85).aspx). TODO: read them some more and edit this line.

IIRC on OSX the process for associating filetypes with applications is the same as setting the icon.

-----------------------------

Once the system knows to use OpenBoardView for `.brd` files, we still need to actually open the file the system gives us. This will likely just be a case of passing `argv[1]` (or `lpCmdLine[0]` if we are still using `WinMain`) to our open file method.

piernov:
It's done for Linux (with GTK) but not for other platforms in R7.3.

piernov:
(sorry, misread that, forget the previous notification)

PiezPiedPy:
> we still need to actually open the file the system gives us. This will likely just be a case of passing argv[1] (or lpCmdLine[0] if we are still using WinMain) to our open file method.

I'll have a look at adding this in a few weeks if you don't mind, its a feature that would make my life a tiny bit easier when opening files 😉

piernov:
Done already a while ago. See OpenBoardView/OpenBoardView@0cc08bd361abeceaf3c9a00bdd59e52ca13557c4.

Issue #58, opened 2016-11-13
Please include .desktop, extensions file and AppStream's AppData


srakitnican:
[openboardview.zip](https://github.com/OpenBoardView/OpenBoardView/files/587771/openboardview.zip)

So this two files are required for packaging. In Fedora at least, GUI applications that are packaged in official repository must have this. I think it makes sense shipping them with sources as it can be reused for other distributions as well. I've attached something I came up with that you are free to use as an example or as is.

openboardview.desktop
```
[Desktop Entry]
Encoding=UTF-8
Name=OpenBoardView
Comment=Software for viewing .brd files
Exec=openboardview -i %f
Icon=openboardview
MimeType=application/x-openboardview-board
Terminal=false
Type=Application
Categories=Utility;Viewer;Electronics;GTK;
```
openboardview.xml
```
<?xml version="1.0" encoding="utf-8"?>
<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
  <mime-type type="application/x-openboardview-board">
    <sub-class-of type="text/plain"/>
    <comment>OpenBoardView PCB Layout</comment>
    <glob pattern="*.asc"/>
    <glob pattern="*.bdv"/>
    <glob pattern="*.brd"/>
    <glob pattern="*.bvr"/>
    <glob pattern="*.fz"/>
  </mime-type>
</mime-info>
```

MattSturgeon:
IIRC I had these files generated by the CMake scripts before merging back inflex's changes. I'm on my phone at the moment, but I'm sure someone can look at getting these files included in the build

srakitnican:
[openboardview.appdata.zip](https://github.com/OpenBoardView/OpenBoardView/files/1156900/openboardview.appdata.zip)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<component type="desktop">
  <id>openboardview.desktop</id>
  <metadata_license>CC0-1.0</metadata_license>
  <project_license>MIT</project_license>
  <name>OpenBoardView</name>
  <summary>Viewer for PCB layouts</summary>
  <description>
    <p>
      Software for viewing PCB/Laptop/Motherboard Layouts.
    </p>
    <ul>
      <li>Dynamic part outline rendering, including complex connectors</li>
      <li>Annotations, for leaving notes about parts, nets, pins or location
      </li>
      <li>Configurable colour themes</li>
      <li>Configurable DPI to facilitate usage on 4K monitors</li>
[...]

Issue #59, opened 2016-11-15
Fileformat: CAST ( .LST plain text / .CST 8-bit )


inflex:
File format to implement, at least this one looks moderately rudimentary being plain text (the .LST parts at least, the CST seems 8-bit)

This package contains a viewer as well, and a pdf

[cast.zip](https://github.com/OpenBoardView/OpenBoardView/files/591816/cast.zip)

piernov:
![capture d ecran_2016-11-15_22-06-21](https://cloud.githubusercontent.com/assets/3323883/20324178/cdd74ed0-ab7f-11e6-8bab-e1e65402f164.png)
Nowhere near being usable but it's a start.

inflex:
Less than 12 hours and you're already at that stage, I'm impressed!

What's left still to do?

piernov:
For the most basic support, I have to figure out which field tells on which side the component is (should not be too hard) and I have to parse the nets list and find the corresponding field in the pins (should be rather easy too). Also the current code is a dirty mess. I still have to find out where the board outline is defined. Then come the test points.
This format also packs in a few extra features not supported in OBV like vias, traces and pin/component shapes.
Like FZ it provides extra information on the parts in the LST file which is not parsed at all for now.

inflex:
A shapes library, with standard outlines is certainly something that needs to be worked on in the future, though at least for now the rendering is "sufficiently ample" for the task at hand, but it's always nice to do better when there's nothing else left to do :+1: 

The part information at least we can currently display like we do with the FZ format.  While it's not separated/parsed in to smaller units (name/value/outline) at least the collective information can still be shown [in the side bar or hover]

piernov:
![capture d ecran_2016-11-16_08-21-10](https://cloud.githubusercontent.com/assets/3323883/20338173/ac60ff16-abd5-11e6-8eec-612a90144633.png)

Looks better. The side of the board is defined by the layer, I need to figure out if there's way to know the number of layers, for now 1 & 12 are hardcoded.

inflex:
Can you process using 1 = top (or bottom), anything > 1 = other side?  Or wait till you know the maximum number of layers and then post-process the data to set the side accordingly?

piernov:
Committed my work here: https://github.com/piernov/OpenBoardView/commit/0b8ce0f394751e2fca0463a991c69b27eee7dd8b
Not finished but I'm not sure I will have the time to do it…

inflex:
[...]

PR #82, opened 2017-07-20
Add asset files for Linux desktop

srakitnican:
Issue #58.

.desktop file slightly modified from original:
- Removed Encoding since it causes a deprecated warning with desktop-file-utils
- Updated Comment based on the same from appdata


piernov:
Gonna have to try that but it looks great. Travis is (hopefully) currently generating the packages on my repo.

MattSturgeon:
Looking at the changes, this might also fix #33

piernov:
It does but only for GNU/Linux. On Windows we still have to make changes to the registry, and for macOS I don't have a clue.

piernov:
CMake-generated RPM and DEB packages available here: https://github.com/piernov/OpenBoardView/releases/tag/N13 . I will take a look at them tomorrow.

MattSturgeon:
Regarding  `make install`s, I think the install target should always act like a full system install, it should be agnostic of the install location (`/`, package-dest or some local directory).

That said you can always add cmake CACHE variables to configure the build, like how within the install location you can also set the prefix (and on some build systems set separate prefixes for different file types, e.g. `/usr/local` for bins and `/usr` for `.desktop`).

There's nothing stopping you making a toggle for "install asset files", though realistically it's probably best to just include them in all installs, incase someone wants to use them for something manually.

piernov:
Yep good point. So it's fine like that. We only have to make sure everthing works as it should.

asset/CMakeLists.txt
```diff
@@ -5,6 +5,15 @@ if(APPLE)
...
 	)
+elseif(UNIX)
+	install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/ofbv-app.svg
+		DESTINATION share/icons/hicolor/scalable/apps RENAME openboardview.svg)
+	install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openboardview.appdata.xml
+		DESTINATION share/appdata)
+	install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openboardview.desktop
+		DESTINATION share/applications)
+	install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openboardview.xml
+		DESTINATION share/mime/packages)
```
MattSturgeon: I think the paths should be `/usr/share/` not `share/`. CMake will prefix relative paths with `CMAKE_INSTALL_PREFIX` (usually `/usr/local` or `/usr`), but these files should always go under `/usr`.

srakitnican: Well if you are installing under `/usr/local`, you would want to install the files under `/usr/local/share`, as well?

srakitnican: https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html#s2_layout

srakitnican: This also makes it possible to install OBV in in $HOME directory for people that can't write to system directories, for example:
make install DESTDIR=$HOME/.local

MattSturgeon: Agreed. Do you have a PR for srakitnican/OpenBoardView@8e5aaac (and any other useful changes)?

Probably should have a separate PR with that change and to refactor `build.sh` slightly to make use of DESTDIR

piernov:
Also for MIME types we have (somewhat working) CAST support (see #59 ).

The other MIME types could be a bit more descriptive. For example `application/x-openboardview-brd-encoded` could be `application/x-openboardview-brd-landrex` or `application/x-openboardview-brd-testlink` or whatever. Although it's still OK like as it is now.

Maybe you could rename the `asset/ofbv-app.svg` to `asset/openboardview.svg` directly in the Git repo? (the `ofbv` part doesn't make sense anymore anyway)

MattSturgeon:
Re the icon name: probably better to do that separately, perhaps a PR to review asset names in general (e.g. should the other icons still be named `brd`)?

I agree it should be renamed, and itd be nice to have it done before this, but it feels kinda out of scope here, especially if we are going to squash merge.

More descriptive MIME names would be nice, realistically if it's including the program name, does it need the extension too (e.g. `application/x-landrex` or `application/x-openboardview-landrex`)? We should probably check any of these filetypes already have "official" mimetypes too.

piernov:
Ok, I was talking about `ofbv-app.png` in particular because @srakitnican explicitely renames it when installing the asset file.

About the MIME type names I've no idea. Maybe we shouldn't bother too much and leave it at that.

MattSturgeon:
Probably best if @inflex and @piernov take a look at the "magic" MIME detection values just to ensure they will work well for all filetypes.

It would be a shame if we check against a string or a byte that isn't always the same (or give false positives from other filetypes)

piernov:
Should probably be matching with https://github.com/OpenBoardView/OpenBoardView/blob/master/src/openboardview/BoardView.cpp#L422 (and the various verifyFormat() methods).
I think I was too lazy with the CST file…

srakitnican:
Collisions are not fatal, this is just a convenience thing, for example if other application claims the same extension with greater weight it can have it, but if matches any magic bits, it should be preferred by OBV. User also gets a nice recommended applications overview in "Open With" menu.

Why `application/x-openboardview-...` so that you know what to erase when it gets all over your files :) There is nowhere to be found specifications for any of these file types, nor what it is called. So I thought the best here is to name it after application that installed this extension. But feel free to change it later to any of the likings.

I am not very familiar with all the file types, so I wrote to the best I could based on source code found here and information found mostly from your posts.

MattSturgeon:
### Overview of filetype matching done by BoardView.cpp and the File classes.

#### using file extension only

FZFile ".fz"
ASCFile ".bom" ".asc"
CSTFile ".cst"

#### Comparing the buffer as hex

BRDFile
  - Check first 32 bits match `{{0x23, 0xe2, 0x63, 0x28}}`

#### using a string search

BRDFile2
  - `find_str("BRDOUT:") && find_str("NETS:")`

BDVFile
  - `find_str("dd:1.3?,r?-=bb")`
    - or
  - `find_str("<<format.asc>>") && find_str("<<pins.asc>>")`

BVRFile
  - `find_str("BVRAW_FORMAT_1")`

#### For comparison, here's how it is checked in mime-info

x-openboardview-board
  - filename "*.{bdv,bvr,fz}"
    - @srakitnican is `{,}` globbing supported for mimetypes? If so it would compact three lines into one.

x-openboardview-asc
  - filename "*.asc"
  - position 8:60 matches "eM-Test Expert"

x-openboardview-brd-landrex
  - filename "*.brd"
  - Check the first 32 bits match `0x23e26328` _(same as code, but with addition of file extension)

x-openboardview-brd-r4
  - filename "*.brd"
  - position 0:30 matches "BRDOUT:"

x-openboardview-lst
  - filename "*.lst"
  - position 3:120 matches "$ASM"

### Thoughts

I notice one of the main differences is that the magic mine checks an exact row:column position, while the C++ implementation does a full string search. While I'm rather unfamiliar with the file formats, it feels like the precision of the row:column check could be prone to breaking for some of the cases.

Also worth checking if there's any filetypes that could be checked better (either here or in our code) based on the other implementation. Again, just raising potential points, not my area of expertise.

---

If all of the above is ok _(or once it has been used for improvements)_ I think this may be ready for a merge.

srakitnican:
> x-openboardview-board
> 
> filename "*.{bdv,bvr,fz}"
> @srakitnican is {,} globbing supported for mimetypes? If so it would compact three lines into one.

I don't think so, at least I haven't seen anyone using it. In fact the only thing I saw is `*`.

> x-openboardview-lst
>
> filename "*.lst"
> position 3:120 matches "$ASM"

If position 0 matches "$VS2"

MattSturgeon:
> > x-openboardview-lst
> > 
> > filename "*.lst"
> > position 3:120 matches "$ASM"
>
> If position 0 matches "$VS2"

Oops, edited.

> the only thing I saw is `*`.

Yeah, `{ }` list globs aren't used/known as much, but that doesn't mean they aren't supported. Not a big deal either way.

---

So are you happy with how the file matching compares between here and obv code? Are there any magic checks that could go? _(filename globs are prefered for performance reasons because the system doesn't have to open the file to do the check)._ Are there any checks that should be added? Should any magic checks by nested under the glob check or vice versa? Are there any checks that are too vague and might match other files we can't read? If so is that an issue?

srakitnican:
Well I have tested only on files attached in issues, and manually generated files of scenarious I can think of. What is there in the wild I have no clue. As I said already it would help tremendously is we would work by some specification, as we can't do that the only thing left is to do it by trial and error. I can't think any reason where false matching or non matching would cause any major issues except annoyance.

> ASCFile ".bom" ".asc"

I didn't add .bom as I've read is not supported currently.

> BRDFile2
> find_str("BRDOUT:") && find_str("NETS:")

Searching only for "BRDOUT:", less searching the better as I have no firm indication of where other string could be, and searching for it in entire file is not a good idea as I've read.

> BDVFile
>
> find_str("dd:1.3?,r?-=bb")

I've missed this one, if you think is neccessary and consistent I could add it

> or
> find_str("<<format.asc>>") && find_str("<<pins.asc>>")

I don't think I can search for this one because file is encoded originally as I understand.

> BVRFile
> find_str("BVRAW_FORMAT_1")

If you think is neccesary I could add this one as well, although searching for it in entire file is not a good idea, if you know for sure of the location(s).

MattSturgeon:
Searches are more costly than looking at a specific case, however they shouldn't be much of an issue if they only happen after checking something cheaper (e.g. nested under the glob).

Still, it's minor nit-picking, we can always open issues in future to improve the file matching both for the mime types and in the code.

@piernov are you happy to squash and merge?

srakitnican:
> Searches are more costly than looking at a specific case, however they shouldn't be much of an issue if they only happen after checking something cheaper (e.g. nested under the glob).

Except they are not nested, try removing the extension, they are still searched for and matched :)

MattSturgeon:
I wasn't saying it _is_ nested, I was saying it possibly could be - if we modified the XML structure.

For example for `"BRDOUT:" && "NETS:"`, you could nest `"NETS:"` under your existing `"BRDOUT"`, but it isn't really needed.

Either way it isn't important ATM, as I said in my last comment, assuming @piernov is happy I think this is probably ready for merge.

srakitnican:
Nesting `<match>` is possible yes, but nesting `<match>` under `<glob>` is not. Yes your point about nesting `NETS:` under `BRDOUT:` shouldn't be too expensive, I guess we could do that if we are sure we can find it there.

MattSturgeon:
> I notice one of the main differences is that the magic mine checks an exact row:column position, while the C++ implementation does a full string search.

Doh! Just read [the spec](https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-0.11.html#idm139839923550176) and it turns out I'm an idiot, the `:` is indicating a range of bytes to search, not an exact `row:column` to check.

I think I'll just merge this since I can think of no reason why it will cause any issues.

piernov:
Seems to work with various FZ/BRD/BDV/ASC (what I've got here right now) so I'm happy with that.

srakitnican:
What we could do for BRD file type if we don't want to associate with any other files is to remove extension association and rely only on magic. That way it will not try to open Cadence file format for example.

PR #86, opened 2017-07-22
CMake: Don't overwrite CMAKE_INSTALL_PREFIX

srakitnican:
Common use case is to pass `DESTDIR` as a destination directory,
and `CMAKE_INSTALL_PREFIX` as a prefix:

```bash
cmake -DCMAKE_INSTALL_PREFIX=/usr
DESTDIR=/tmp/root  make install
# Not sure how DESTDIR behaves with `make package`
```

Without this change, this would result in unwanted prefix. And
there was no way of controlling install directory externally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,8 +50,6 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 	endif()
 endif()
 
-# Install in project directory
-set(CMAKE_INSTALL_PREFIX ${PROJECT_SOURCE_DIR})
 # Extra CMake modules (Find*.cmake) are in CMakeModules/
 set(CMAKE_MODULE_PATH
 	${CMAKE_MODULE_PATH}
--- a/build.sh
+++ b/build.sh
@@ -26,7 +26,8 @@ if [ "$1" == "--help" ]; then
 fi
 STRCOMPILE="$(tput bold ; tput setaf 2)Compiling$(tput sgr0)"
 COMPILEDIR="release_build"
-COMPILEFLAGS=""
+COMPILEFLAGS="-DCMAKE_INSTALL_PREFIX="
+export DESTDIR="$(cd "$(dirname "$0")" && pwd)"
 BUILDTYPE="$(tput bold ; tput setaf 6)release$(tput sgr0)"
 SCRIPT_ARGC=1 # number of arguments eaten by this script
 if [ "$ARG_LENGTH" -gt 0 -a "$1" == "--debug" -o "$2" == "--debug" ]; then


MattSturgeon:
+1

Needs modifications to `build.sh` though to avoid installing system wide by default. Some way to set the prefix and the install location (DESTDIR) and have default values.

I think the script was written by @inflex

CMakeLists.txt
```diff
@@ -51,7 +51,9 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif()
 
 # Install in project directory
-set(CMAKE_INSTALL_PREFIX ${PROJECT_SOURCE_DIR})
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+	set(CMAKE_INSTALL_PREFIX ${PROJECT_SOURCE_DIR})
+endif()
 # Extra CMake modules (Find*.cmake) are in CMakeModules/
```
MattSturgeon: Personally I'd vote for not setting it at all. I'm happy to leave the prefix as default and instead set a DESTDIR value in `build.sh`.

I think it causes more issues than it fixes when you are trying to second guess what the user wants when they leave he prefix as default, to me overriding even the default is a bad idea.

srakitnican: I was thinking that as well, but I am not sure how would that work for other platforms. In particular I am referring to this note:

> WARNING: DESTDIR may not be used on Windows because installation prefix usually contains a drive letter like in “C:/Program Files” which cannot be prepended with some other prefix.

https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html

MattSturgeon: Currently the build doesn't work on Windows anyway, but even then an user can just set the prefix themselves if they want to.

Most of the time it will just be using cpack (`make package`) to generate zips or installers for Windows anyway.

Personally I'd prefer to have the variables set correctly

piernov:
Moving the prefix setting to `build.sh` seems like a better idea indeed, although in most cases packagers will be explicitely setting `-DCMAKE_INSTALL_PREFIX` in their build script and end users won't try to mess with it.

Also I tried to fix building on Windows with CMake/MSVC a while ago, but I guess it may be broken again.

@MattSturgeon I'm the one who wrote all the stuff related to building/packaging so I'm the one to blame for this. :p

MattSturgeon:
The way I see it, `buils.sh` is an opinionated wrapper on the configurable cmake system.

Going with that principal, distros and packaging scripts should be using CMake directly and setting variables manually and guesswork and opinions like setting DESTDIR or prefix should be moved into the wrapper script.

---

Also didn't realize you'd fixed the windows build, not looked at it since you last said it was broke.

srakitnican:
> Not sure how DESTDIR behaves with `make package`
```
# make package DESTDIR=/tmp/OBVpkg/
...
CPack: - package: /builddir/build/BUILD/OpenBoardView/OpenBoardView-R7.4-Linux.rpm generated.
# rpm -qpl /builddir/build/BUILD/OpenBoardView/OpenBoardView-R7.4-Linux.rpm
/usr/bin/openboardview
/usr/share/appdata
/usr/share/appdata/openboardview.appdata.xml
/usr/share/applications
/usr/share/applications/openboardview.desktop
/usr/share/icons
/usr/share/icons/hicolor
/usr/share/icons/hicolor/scalable
/usr/share/icons/hicolor/scalable/apps
/usr/share/icons/hicolor/scalable/apps/openboardview.svg
/usr/share/mime
/usr/share/mime/packages
/usr/share/mime/packages/openboardview.xml
```
CPack uses its own prefix, "/usr" by default, controllable with CPACK_PACKAGING_INSTALL_PREFIX .

MattSturgeon:
> CPack uses its own prefix

Wonderful.

As I suggested before, I think this PR should be taken a step further, completely removing setting `CMAKE_INSTALL_PREFIX` from `CMakeLists.txt` and instead handling (both?) prefix and DESTDIR withing `build.sh`.

If you're not comfortable working on `build.sh` feel free to change your commit to just cleaning up CMakeLists and then we can look at fixing build.sh after.

/cc @OpenBoardView/maintainers any objections or thoughts?

inflex:
Am watching the work being done; just not my field of expertise - very glad to have those who know what they're doing here. Deferring to others :+1:

srakitnican:
Alright, I hope I did it right, I am not very familiar with git.

MattSturgeon:
> Alright, I hope I did it right, I am not very familiar with git.

I think piernov ment we would do the squashing, but we'll done none the less!

Merged

PR #87, opened 2017-07-23
Drop Dx9 support and standalone VS project (closes #85)

### LINKED ISSUE #85: Drop DirectX 9 support
We still have `main_dx9.cpp` and `imgui_impl_dx9.cpp` lying around, even though they are not used anymore for the official Windows build (since we cross-compile the OpenGL stuff). Should we delete them from the repository since they can be considered as dead code, or does anyone see any use for them?

piernov:


MattSturgeon:
So after this there's still three separate renderers? Still, it's a step in the right direction. Looks good.

piernov:
They are all OpenGL + SDL2-based so it doesn't add much to the maintenance. OpenGL 3.2 is the default one (well… except that there was a mistake in the default configuration file), OpenGL 1.1 is here for backward compatibility with older GPUs (AMD/NVidia < 2007, Intel < 2011), and we've got OpenGLES 2.0 for embedded devices (Android for example).

MattSturgeon:
I'd say it's worth considering how much benefit we get from each GL version. It's probably not possible to go down to only one, but maybe there are some potential compromises. 

- Does GL3.2 give us any noticeable benefits over either of the others?
- Does GL1.1 work on android or does it have to be GLES?
- IIRC, GLES2.0 is pretty new, so probably can't replace GL1.1 if we want to support older GPUs. How does it compare against GL3.2?

MattSturgeon:
Either way, imo this is ready for a merge

piernov:
GL 3.2 at least uses shaders, which are way more efficient on newer GPUs.
GL 1.1 will not work on Android, and the fixed-function path used there is deprecated. On newer drivers it is even only a "software" implementation using shaders.

A better compromise would have been OpenGL 2.1 which makes use of shaders and works on older hardware. However this specific version is not integrated in ImGui, and the support of OpenGL 2.1 on C2D-era Intel integrated graphics is really poor.

GLES 2.0 is a stripped down version of GL 2.0, for embedded devices. On Android for example, regular OpenGL is not supported, only OpenGLES is. It doesn't really make sense to use it on desktop, but could be used if the GPU has OpenGLES 2.0 compatibility and not OpenGL 3.2. I origally integrated it for the Android port.

PR #88, opened 2017-07-24
Build.sh: Cleanups

srakitnican:
I have found the script very difficult to follow with all text formating code that was around, there is still some text formatting code, but much smaller and IMO much better. Also moved help text from main code to a function in the beginning of the script.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -6,6 +6,18 @@ fmtText() {
   echo "$(tput bold; tput setaf ${color})${text}$(tput sgr0)"
 }
 
+helpMsg() {
+  cat << EOH
+Usage: $(fmtText 4 ${0}) [--$(fmtText 5 recompile)] [--$(fmtText 1 debug)] — Build $PROJECT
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Looks nice. Haven't tested it out, but I agree the script could do with some clean up.

I think `fmtText` could possibly be better named. `"$(color 2 Compiling)"` might read slightly better, for example.

srakitnican:
I've put format originally because we are also changing font weight and whatever else could happen in that function in the future. But I don't have a strong opinion on that as long as the code is readable.

MattSturgeon:
I like `fmt` and `color`. Either of those is fine, but `fmt` might need a comment explaining what it is/how it works since it is kinda ambiguous.

MattSturgeon:
Merged as 2f7c67182251a57f0145100e944930e0921d10ca and a97bb92cab93a69c8acc7f433f8444ef01d91af6

PR #89, opened 2017-07-25
bvconv: Fix it so that it produces readable files

srakitnican:
According to what is OBV looking for, this was not producing a
valid file.
- Spelling in header
- [Layout] to <<Layout>> etc.
- Tab as column separator

### DIFF
--- a/utilities/bvconv.sh
+++ b/utilities/bvconv.sh
@@ -12,11 +12,11 @@ echo "Usage: $0 input.bv > output.bvr"
 exit 1
 fi
 
-echo "BRRAW_FORMAT_1"
-echo "[Layout]"
-mdb-export "$1" Layout 
-echo "[Nail]"
-mdb-export "$1" Nail
-echo "[Pin]"
-mdb-export "$1" Pin
+echo "BVRAW_FORMAT_1"
+echo "<<Layout>>"
+mdb-export -d$'\t' "$1" Layout
+echo "<<Nail>>"
+mdb-export -d$'\t' "$1" Nail
+echo "<<Pin>>"
+mdb-export -d$'\t' "$1" Pin
 


srakitnican:
Also, I would like to know if outputting to stdout `Usage: $0 input.bv > output.bvr"` is useful in any case.

I am in the process of re-writing it with an option to just generate the output file along with the input file, different extension obviously. Does it makes sense to keep this functionality as is?

piernov:
Take a look at https://github.com/inflex/bv2bvr instead.

srakitnican:
@piernov What about it? I saw that, but is not merged? Is that meant to integrate within OBV so that there is not need for a user to do conversion?

I still think this one needs to be fixed(or replaced) as is currently not working.

piernov:
@inflex will give you more details about this (as I still did not understand what he wanted to do with this new tool).

inflex:
The reason the tool was spun off was to avoid potential GPL/licence issues. 

I was scratching my head why the shell-script wasn't producing the right output as per the previous posts, then I remembered it was because I dropped that at some point to produce the dedicated bv2bvr conversion tool outside of the project for the above mentioned reasons.

piernov:
There will be no license issue as long as you only use the library part of `mdbtools`. This specific portion of the source code is released under LGPL, not GPL, which means you can link it dynamically with MIT-licensed software. The rest of `mdbtools` source code is under GPL, this would require the software to be licensed under GPL as well.

inflex:
Likely so - I'm just a paranoid old chap.   What my plan was in the end, was to put bv2bvr behind a website and have people submit their .bv files and have them converted in to .bvr on the fly; the upside being you'll also acquire a collection of boardview files ;)

piernov:
I personally hate web-services but well, could do the job… The thing is, as it is right now, either tools can only be used on Linux, manually installing mdbtools.

If we use your C tool, we could probably build the mdbtools library for other operating systems and integrate it as a part of OpenBoardView build process.
In fact, we would even be able to integrate the source code directly into OpenBoardView and do the conversion automatically when opening a file.

The downside is to add glib as a dependency, already required for FontConfig & GTK on Linux, but not on other systems. But there is no other way to make BV files work.

inflex:
I'd prefer not to integrate the code in to OBV given the extremely limited number of BV files encountered and the ease in which they are converted already, particularly given the complexities that would come with it as you've mentioned regarding having glib as a dependency.  

Could be somewhat easier and maybe more useful to the greater community to instead have someone code up a portable/clean-source replacement the reading of the AccessDB format.

piernov:
I mean, currently they cannot be opened on Windows and macOS so it doesn't make any sense to say that it is supported in OpenBoardView (most people use Windows), and a web-service is really not the right solution to the problem in my opinion.
Writing another library doesn't really make sense either, except to get rid of the glib dependency, and it would require too much work. glib isn't that heavy in the end so I don't take it as an issue to add it as a dependency on Windows. macOS people wouldn't even notice it since they're used to alarm clock apps eating up a couple hundred MB of disk space.

srakitnican:
Alright, but can we fix bvconv for now?

I was thinking of doing something like following [bvconv.desktop-demo.zip](https://github.com/OpenBoardView/OpenBoardView/files/1173517/bvconv.desktop-demo.zip) but that is Linux only and I don't do C, sorry.

MattSturgeon:
I'd personally prefer to end up with one tool. I don't mind if it is a util here, a separate repo or built into OBV.

If we go with a separate repo, it would be nice to maybe move [bv2bvr] to the [OpenBoardView](/OpenBoardView) group and start doing cross platform binary releases like we do here, then link to it from the OBV README or something. Could even have a popup in OBV talking about the other tool if someone opens a BV file.

I assume from the above discussion that bv2bvr can be cross-compiled for osx and win32?

Either way if we are moving towards bv2bvr or a built in solution I'd prefer to drop bvconv entirely. I don't see the point in duplicating efforts here.

[bv2bvr]: /inflex/bv2bvr

piernov:
`bv2bvr` cannot be cross-compiled as of now. I think `mdb-tools` itself cannot. A few months ago I planned to take a look at writing a CMake script to built the library and the tool. I'll try to do that in the next days.

For now I'm OK with merging the fix, we'll see in the future how we manage to integrate the conversion.

MattSturgeon:
ok, probably best to open a separate issue to track the future of bv conversion. I'll merge this as you wish.

Issue #93, opened 2017-08-03
Screensaver and Sleep disabled when OBV is open on OSX


tiagobaracho:
IF OBV is open on OSX ( in my case sierra) it makes screensaver never came up nor the sleep / monitor off.... I have to close OBV to screensaver and sleep back to work.
Thanks guys .

inflex:
I better get my macbook setup with the build tools.

I'll be curious to see what's causing this. Might need to use something other than usleep() perhaps on the macOS to enable sleeping, or maybe it's the SDL polling (even SDL_waitevent() isn't any good, as it seems to be just a wrapper around the same polling, as opposed to true event-driven ).

MattSturgeon:
Maybe OSX is treating us as a game/video due to the opengl rendering and not wanting to interrupt the user playing/watching their media?

inflex:
That's a plausible option.  I'll set it up and see what's happening with the main loop, and I'll dig around to see if there's any talk about SDL/OpenGL keeping macOS awake.

piernov:
```
Hikari:~ piernov$ pmset -g assertions
2017-08-03 16:52:59 +0200 
Assertion status system-wide:
   BackgroundTask                 1
   ApplePushServiceTask           0
   UserIsActive                   1
   PreventUserIdleDisplaySleep    1
   PreventSystemSleep             0
   ExternalMedia                  1
   PreventUserIdleSystemSleep     1
   NetworkClientActive            0
Listed by owning process:
   pid 217(nsurlsessiond): [0x0000012b000b80f0] 00:02:19 BackgroundTask named: "NSURLSessionTask F61AE2A7-8250-440D-8689-93C6BD00F213" 
   pid 217(nsurlsessiond): [0x000000d7000b80e3] 00:03:43 BackgroundTask named: "NSURLSessionTask 37E1827B-DCE7-4EF3-BE49-04812D3489F6" 
   pid 217(nsurlsessiond): [0x000000cf000b80dd] 00:03:51 BackgroundTask named: "NSURLSessionTask 7BA1B1FC-74A0-4D43-9901-EC76A0FAD687" 
   pid 566(openboardview): [0x00000099000580a1] 00:04:45 PreventUserIdleDisplaySleep named: "openboardview using SDL_DisableScreenSaver" 
   pid 44(powerd): [0x0000002a00088000] 00:06:37 ExternalMedia named: "com.apple.powermanagement.externalmediamounted" 
[...]

PR #94, opened 2017-08-04
Force SDL not to disable screensaver

piernov:
Dunno if it works but sending this out as a PR anyway. Seems like the issue #93 completely disapperead…

Anyway, test builds here for anyone who wants to try out: https://github.com/piernov/OpenBoardView/releases/tag/N15

### DIFF
--- a/src/openboardview/main_opengl.cpp
+++ b/src/openboardview/main_opengl.cpp
@@ -360,6 +360,9 @@ int main(int argc, char **argv) {
 	cleanupAndExit(1);
 #endif
 
+	// SDL disables screen saver by default which doesn't make sense for us.
+	SDL_EnableScreenSaver();
+
 	ImGuiIO &io    = ImGui::GetIO();
 	io.IniFilename = NULL;
 	//	io.Fonts->AddFontDefault();


inflex:
Thanks @piernov .  Will check it later when I'm done with today's work.

src/openboardview/main_opengl.cpp
```diff
@@ -360,9 +360,11 @@ int main(int argc, char **argv) {
 	cleanupAndExit(1);
 #endif
 
+	// SDL disables screen saver by default which doesn't make sense for us.
+	SDL_EnableScreenSaver();
+
 	ImGuiIO &io    = ImGui::GetIO();
 	io.IniFilename = NULL;
-	//	io.Fonts->AddFontDefault();
```
MattSturgeon: Being nit-picky (as always!) but wouldn't this be better off in a separate cleanup commit? Probably ok to keep it in the PR though.

piernov:
Any feedback on this?

inflex:
Haven't had a chance to test yet @piernov , been too busy :(

srakitnican:
I've tested binary from RPM package. It does not inhibit screen blanking on Fedora 25.

piernov:
The issue was showing up mainly on macOS, Windows was affected when the window was focused. Not sure if it was doing anything on Linux.

srakitnican:
Well, R7.3 that I've packaged for Fedora inhibits screen blanking.

MattSturgeon:
Sounds like the fix works then, given the same setup can reproduce the issue without this PR but can't with it.

PR #95, opened 2017-08-10
CPack: rpm: Don't own directories from filesystem

srakitnican:
Directories excluded in this patch are owned by other packages which CPack adds automatically if not excluded. Common practice is to own only files and directories which are added by package and are not owned by other package already. Mentioned directories are in turn owned by:
`filesystem`, `hicolor-icon-theme` and `shared-mime-info` packages on Fedora. There is at least `filesystem` package on SUSE with same functionality.


MattSturgeon:
Merged as 1516d5172b45a17f78b4e9502296fe64968b203b

PR #98, opened 2017-08-14
Android support

piernov:
Not expecting this to be merged. Just for information & (mostly) fun.


danieleftodi:
@piernov I saw that in your Repro, do began on an iOS Support to :-) 

How's that going?

piernov:
It was 2 years ago so I don't remember exactly, I think I got it to build and open on emulator, but there was still a lot of issues. Especially there is no file management on iOS so I don't know how to deal with boardview files.

danieleftodi:
> It was 2 years ago so I don't remember exactly, I think I got it to build and open on emulator, but there was still a lot of issues.

That is kinda like half-way there, which is good :-) 

> Especially there is no file management on iOS so I don't know how to deal with boardview files.

...ahh, yeah ... two years ago ... you didn't have the options you have now :-) 
... they implemented something more towards "Android" ... so it should be easier to implement "a filesystem" now in iOS 14 ... 

If I get some spare time over this summer, I'll try to get it to run from where you left it be ... and see what I'll be able to hack together. 

Would be nice to have a Boardview app on an iPad Pro, and such :-) 

BTW, I looked into the .brd files ... and recall I used Eagle to open them / make them when tinkering with my own board designs ... But now when I tried to open them in the "Newer Eagle" (After Autodesk bought them ...), I complains that it is not an XML-format file (It cannot find the start xml-tag...)

Do you know if an older version of Eagle, or something else, can open these binary .brd files .... and export/save (convert) them to the newer XML format in Eagle?

Cheers,
Daniel

PR #99, opened 2017-08-15
Use cmake templating to inject project variables

MattSturgeon:
In order to centralise things like project name and reduce hardcoded values, we should inject CMake variables into the code (`version.h`) and relavent assets. This could be easily extended further in future, perhaps also setting description strings and other gneric metadata.

I know there was some resistence to this idea in the past, hence PR. IMO this improves maintainablity without adding undue complexity. Thoughts @OpenBoardView/maintainers, @srakitnican?


srakitnican:
For the part about assets.

Referencing file names from sources tree with variables is having the opposite effect IMO since changing variable does not magically change filename as well. For example; if you want to temporarily change project name you will get errors on missing files. I can understand for generated files, but this does not make any sense:

```cmake
	install(FILES "${PROJECT_NAME_LOWER}.svg"
	        DESTINATION "share/icons/hicolor/scalable/apps")
```
Also noticed that you mix indenting character on various places between tab and space.

Something maybe worth considering for the future; Not sure if there is some kind of changelog now, but it would be nice to implement something that can update AppData file as well. Not sure if that is possible just with CMake in a sane way, though.
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

MattSturgeon:
> Also noticed that you mix indenting character on various places between tab and space.

That was intentional. The indentiation is using tabs, but aligning function arguments over multiple lines uses spaces, since tabs could be any width and won't align with othe characters.

> Referencing file names from sources tree with variables is [bad because] changing variable does not magically change filename

Agreed. I'd prefer to avoid having the assets use the name in the repo and then use `install()`'s `RENAME` option or rename using `configure_file()`. For now it's probably best to keep it hardcoded when it refers to an actual filename.

For the specific example you quoted, I'd suggest a pattern like:

```cmake
install(FILES "icon.svg"
        DESTINATION "share/icons/hicolor/scalable/apps"
        RENAME "${PROJECT_NAME_LOWER}.svg")
```

> Not sure if there is some kind of changelog now, but it would be nice to implement something that can update AppData file as well

Hm, certainly out of scope for this PR, but I haven't really thought about how best to approach changelogs yet.

MattSturgeon:
Applied feedback from @srakitnican: now assets have generic names on disk and are renamed to PROJECT_NAME by CMake. Since I haven't seen any major objection to this, I'll merge now.

PR #101, opened 2017-09-01
MSVC build fixes.

PiezPiedPy:
A number of small changes to allow MSVC to build without compile errors.
See the individual commits for details.


piernov:
I've already mostly fixed the issue here: https://github.com/piernov/OpenBoardView/commits/wip/fix/msvc-2
Except that the win32 `lookup_file_insensitive` doesn't work and I didn't have the time to take a look at it. (it's mostly an SO copy/paste so obviously it wouldn't work)

PiezPiedPy:
Sweet, I'll take a clone of your fork and give the branch a try 😉

MattSturgeon:
@piernov if your changes fix stuff, we should probably get it merged. Maybe without `lookup_file_insensitive` for now?

Surely a build that fails to correctly handle case-insensitive filesystems is better than no build at all. I wonder if any of this will be simplified by C++17's filesystem library...

piernov:
Well, `lookup_file_insensitive` isn't really relevant on case-insensitive filesystem anyway, so for most use-cases on Windows it's useless. I didn't take a good look at it but the issue comes from string encoding conversions (again…). I will still send a PR as-is, not to be merged.

Yeah I'm kind of impatient about that C++ FS library.

PR #105, opened 2017-09-12
Remove the dependency on bash

RyuKojiro:
This allows `build.sh` to completely build OBV on systems that don't have bash, and It'll still work on systems that do. I've tested this on NetBSD 7.1, OS X 10.12.6, and Debian 8.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -75,20 +75,20 @@ fi
 echo "$STRCOMPILE $PROJECT using $(color 4 $THREADS) $STRTHREADS ($BUILDTYPE build)"
 echo "Extra flags passed to CMake: $COMPILEFLAGS"
 cmake $COMPILEFLAGS ..
-[[ "$?" != "0" ]] && color 1 "CMAKE FAILED" && exit 1
+[ "$?" != "0" ] && color 1 "CMAKE FAILED" && exit 1
 if `echo "$COMPILEFLAGS" | grep -q "DEBUG"`; then
   make -j$THREADS install
-  [[ "$?" != "0" ]] && color 1 "MAKE INSTALL FAILED" && exit 1
+  [ "$?" != "0" ] && color 1 "MAKE INSTALL FAILED" && exit 1
 else
   make -j$THREADS install/strip
-  [[ "$?" != "0" ]] && color 1 "MAKE INSTALL/STRIP FAILED" && exit 1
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -43,29 +43,31 @@ COMPILEFLAGS="-DCMAKE_INSTALL_PREFIX="
...
   STRCOMPILE="$(color 5 Recompiling)"
   rm -rf $COMPILEDIR
   SCRIPT_ARGC=$((SCRIPT_ARGC+1))
 fi
-if [ "$CROSS" == "mingw64" ]; then
+if [ "$CROSS" = "mingw64" ]; then
   COMPILEFLAGS="$COMPILEFLAGS -DCMAKE_TOOLCHAIN_FILE=../Toolchain-mingw64.cmake"
 fi
-COMPILEFLAGS="$COMPILEFLAGS ${@:${SCRIPT_ARGC}}" # pass other arguments to CMAKE
+SUBSTRING=$(echo $@ | cut -c${SCRIPT_ARGC}-)
```
piernov: Isn't it supposed to be `cut -f`? Not sure.

RyuKojiro: I was confused about that, too. I think the original implementation intended for it to separate by argument, not character, but bash's `${parameter:offset}` syntax is character indexed. Here is the documentation: https://www.gnu.org/software/bash/manual/bashref.html#Shell-Parameter-Expansion

Now is a reasonable time to change it to `-f` if we want to correct the behavior. Want me to drop a commit on top or fix it up?

piernov: This documentation explains that the behaviour is different for `$@`. This is this specific behaviour that I expected when I wrote this piece of code, which would match with `cut -f` rather than `cut -c`.
However, command line parameters with spaces enclosed in `""` are treated as a single parameter with `$@`, but with `cut -f` they would be separated. In our case it's not an issue, but that's an important thing to remember.

RyuKojiro:
You're absolutely right! I didn't know it behaved differently when used with `$@`.

I've fixed up the `cut` change to use `-d ' '` and `-f`

piernov:
Should be ok, thanks!
Also, are you planning on making a FreeBSD port?

RyuKojiro:
I wasn't planning on it, but it's not a bad idea.

PR #106, opened 2017-09-19
Warning cleanup

Nazg-Gul:
Some commits with no-functional-changes related on silencing strict compiler warnings.

Those warnings are often helping catching real issues in the code, and those are something we keep enabled in Blender. Not sure what is usually policy to deal with such warnings, just used Google style for that. Again, something we use in Blender and some other related projects.

Did some cleanup to help keeping my terminal outputs clean while looking in some other bug.

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -1156,7 +1156,7 @@ void BoardView::ShowInfoPane(void) {
 
 	if (m_partHighlighted.size()) {
 		ImGui::Separator();
-		ImGui::Text("%d parts selected", m_partHighlighted.size());
+		ImGui::TextUnformatted((std::to_string(m_partHighlighted.size()) + " parts selected").c_str());
 
 		for (auto part : m_partHighlighted) {
 


piernov:
I'm not really fond of trying to silence unharmful warnings at all costs so I'm not sure about the first 2 commits. By the way the second one should be commenting the `short shape =` part rather than adding the workaround below. Would love to hear someone else's opinion.
Thanks anyway, and the last commit is ok although the define may be may be added to `platform.h` instead for example.

Nazg-Gul:
I agree it is harmless, but here is motivation:

- It's not crappy of workaround to silence those. There are more crappy ones caused by compiler bug/stupidity.
- Not having warnings in the compile logs helps identifying problems in new code much quicker: you don't need to spend time sorting out already existing warnings and warnings introduced by new code.
- Not having warnings helps using `-Werror`, so you never miss something  important when doing things like refactor.
- In C++ you can wipe unused arguments in a way `void foo(cons char*)`. But this makes it hard to get semantic, so what some styleguides are suggesting is to always use argument name in definition, but just comment it out. This way you always know what to uncomment to get your argument.
- Unused variable is not always harmless from the performance point of view. One might have `int unused_var = someComplexFunctionToCompute()`. This could be both residue from debug and/or from optimization/refactor. In any case, you don't want those. And compiler wouldn't be able to optimize that out reliably, because it might consider that function modifies global state (via `errno` for example).
- Helps during cleanups. You can quickly grep and see all declarations. If you see argument is never used, you can wipe it from the functions completely. Saving some code (sometimes quite a bit of code, due to noodle of passing some weird context), making it easier to follow.
- There are more dangerous warnings, like shadowing which are more important to look into than this particular one. But where to draw a line of a dangerous warning vs. not dangerous one? And why not to address all simple ones to avoid possible nasty things in the future?

Surely, all the above is crucial in big projects and doesn't matter that much here. Just found it a good practice to be always strict. Avoids accidents, and solves problems like "hey, is it a small enough project that I might untighten some nuts in my mind when writing new code?".

Will update patch for shape and `platform.h` in a bit.

Nazg-Gul:
I've force-pushed the branch now. Seems updated commits were picked up.

Let me know if it'll be more convenient if i split this pull request into three different ones, so good/safe parts can be applied to master, but keep conversation open about more controversial changes (i do understand different projects have different policies, i'm jut bringing some knowledge/etc which was proven helpful in projects i'm working on full/part time).

MattSturgeon:
Not sure how I feel about this. I agree we should do what we can to remove warnings and errors, so that any new warnings get the attention they deserve, however commenting out variables seems like a messy solution IMO.

While I appreciate being warned of unused variables may be useful, I'd personally sooner silence the warning and allow the compiler to optimize away unused variables (assuming this is possible/default behaviour in modern compilers).

---

The `size_t` formatting thing makes sense, although I feel like there may be a more C++y approach like using `std::to_string()`. Just a shame you have to convert back to c_strings for most APIs, so it feels inefficient to introduce `std::string`s just for a string conversion/concatenation, even though it probably isn't:

```c++
ImGui::Text((std::to_string(m_partHighlighted.size()) + " parts selected").c_str());
```

Nazg-Gul:
it is probably safe-ish to use `-Wno-unused-parameter`, but then you're increasing risk of not noticing unused argument which causes nasty chain of variables being passed from a parent scopes. Other consideration here is to simple omit argument name in such function, but then the concern becomes that if someone will want to work on that function it will be unclear what that argument actually mean.

Here are some examples [there](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions).

`-Wno-unused-variable` might be dangerous, because you might forget to remove some heavy call which was only used for debugging.

But this is really your call, i'm just sharing knowledge and experience from wide variety of other projects i'm working on.

As for `std::to_string()`, it is C++11 and i didn't see other places where it is required, and didn't want to raise requirement for the compiler. If you think it's fine to require C++11 can easily update the patch or you can just commit own version :)

piernov:
We're actually already using C++11 so it's not an issue. Or at least the parameter `-std=c++11` is passed to the compiler, but I'm pretty sure I used some C++11-only stuff somewhere…
Indeed, I prefer using the C++ way that @MattSturgeon suggests. However `ImGui::Text()` is a `printf`-like function so the first argument must be a formatting string literal (`"%s"` for example) so that doesn't make much sense. There is an `ImGui::TextUnformatted()` function though.

Nazg-Gul:
@piernov, you're right. So `std::to_string()` is fully safe to be used.

So we can do `ImGui::TextUnformatted(std::to_string(m_partHighlighted.size()) + " parts selected")`.

Or screw it all, keep things simple and do `ImGui::Text("%d parts selected", (int)m_partHighlighted.size());`. Will anyone really ever have that many parts on a schematic?...

piernov:
I don't like C-style cast, but a `static_cast<int>(m_partHighlighted.size())` would do the job indeed.

Nazg-Gul:
Committed cast for now. `std::to_string()` turns out to be something like:

  ImGui::TextUnformatted((std::to_string(m_partHighlighted.size()) + " parts selected").c_str());

Can not make it more clean that that. Any opinion if that's indeed better? :)

Nazg-Gul:
Committed the to_string() as well now.

Hoped Github will show versions of the patch, appears it does not (sorry, mainly familiar with Gerrit and Phabricator's code review tools).

Sorry for the noise :)

MattSturgeon:
Github can show individual commit diffs, and can squash them down when
merging if desired.

It's a shame we can't just pass a std::string to imgui, but given it needs
a format string literal, casting is probably the cleanest approach.

Regarding -Wno-unused-parameter, I feel like I'd need to see some examples
of potential pitfalls to decide if it really is worth keeping the warning,
but I'm leaning towards just ignoring it. Thoughts?

Nazg-Gul:
@MattSturgeon, ok, getting a bit confusing. Let's clarify something..

> It's a shame we can't just pass a std::string to imgui, but given it needs a format string literal, casting is probably the cleanest approach.

Casting as in `static_cast<int>` or vector size?

`ImGui` doesn't need any format literal if you use `TextUnformatted`. Not sure why/whether it's important to keep using "formatting" API. It seems to be less expensive to do string concatenation and get `c_str` (which simply returns pointer, without any extra logic as far as i know) rather than go via trouble of `va_list`, and everything..

> Regarding -Wno-unused-parameter, I feel like I'd need to see some examples of potential pitfalls to decide if it really is worth keeping the warning, but I'm leaning towards just ignoring it. Thoughts?

The pull request now doesn't have anything related on unused parameters or variables. It's really minor thing so probably easier just to ignore it.

If the question is whether to add `-Wno-unused-parameter` to CFLAGS via CMakeLists -- then thinking it's not something what should be done. If someone is annoyed by those warnings he can silence those by passing `CMAKE_CXX_FLAGS` argument to CMake.  If that's warning is forced disabled via CMakeLists then there is no way to "resurrect" those warnings without modifying CMakeLists.

MattSturgeon:
It's been a while since I looked at this and I can't recall all the changes from the original commits, however looking at the current proposed change, I think it is a simple question of warnings vs readability.

`Text("%d parts selected", m_partHighlighted.size())` is a lot more readable than `TextUnformatted((std::to_string(m_partHighlighted.size()) + " parts selected").c_str())` but as it's only one line and it is still readable enough, I'll merge this for the benefit of anyone who cares about those warnings.

inflex:
I'd prefer it stay the old way.  Code base is already getting pretty large and making these sorts of changes for the sake of squelching the compiler warnings is (imo) getting a bit too pedantic and hurts the readability (and servicabiltiy).

PR #107, opened 2017-09-19
Fix crash when clicking on CLEAR before loading board

Nazg-Gul:
Not as if it's something what people usually do, but it's easy enough to click by accident.


piernov:
Thanks!

PR #108, opened 2017-09-21
Fix build with MSVC

piernov:
### Do not merge
`lookup_file_insensitive` not working, breaks ASC support.

### DIFF
--- a/src/openboardview/win32.cpp
+++ b/src/openboardview/win32.cpp
@@ -1,6 +1,7 @@
 #ifdef _WIN32
 
 #include "platform.h" // Should be kept first
+#include "utils.h"
 #include "imgui/imgui.h"
 #include "utf8/utf8.h"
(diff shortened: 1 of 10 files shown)


src/openboardview/win32.cpp
```diff
@@ -158,6 +159,26 @@ const std::string get_user_dir(const UserDir userdir) {
 	return configPath;
 }
 
+// Case insensitive lookup of a filename at the given path
```
MattSturgeon: Why do we need case-insensitive on win32? Isn't it by default?

piernov: It can be made case-insensitive. But IIRC I mostly tried to add it to keep the same interface and behaviour than on Linux and macOS, as it was already used before on all platforms but with the POSIX interface (that's not really liked by MSVC).

PR #111, opened 2017-10-04
BRDPart::type refactoring

martonmiklos:
- Renamed BRDPart::type to BRDPart::mounting_side
- Introduced enum for BRDPart::mounting_side

There is a bit hardly understood-able code around the BRDBoard::BRDBoard(const BRDFile *const boardFile), what I have not touched, but I can clean it up to if required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ imgui.ini
 *.tags
 .idea/
 tags
+
+# Do not track QtCreator imported project files
+OpenBoardView.config
+OpenBoardView.files
+OpenBoardView.includes
--- a/src/openboardview/BRDBoard.cpp
+++ b/src/openboardview/BRDBoard.cpp
@@ -87,20 +87,15 @@ BRDBoard::BRDBoard(const BRDFile *const boardFile)
 			if (is_prefix(kComponentDummyName, comp->name)) comp->component_type = Component::kComponentTypeDummy;
 
 			// check what side the board is on (sorcery?)
-			if (
-					(brd_part.type == 1)
-					||(brd_part.type >= 4 && brd_part.type < 8)) {
+			if (brd_part.mounting_side == BRDPartMountingSide::Top) {
 				comp->board_side = kBoardSideTop;
-			} else if (
-					(brd_part.type == 2)
-					||(brd_part.type >= 8)
-					) {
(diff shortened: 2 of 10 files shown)


piernov:
Thanks for the contribution!
A few comments:

- use `clang-format` to fix the formatting after making changes
- `part.type` also describes the kind of part this is (through-hole or SMD), there should be another field with another `enum` to encapsulate this information
- if the purpose is to encapsulate the mounting side rather than using a bare `int`, the `enum BRDPartMountingSide` shouldn't reflect those integers since they are `BRD` specific.
- `enum BRDPartMountingSide` should be an `enum class BRDPartMountingSide`, and the elements named without `MountingSide`, so you could write `BRDPartMountingSide::Bottom` for example. This makes for a better encapsulation too.

martonmiklos:
First of all, many thanks for the review!

So if understand correctly you would like to see two separated members for each BRDPart: a mounting_side (top/bottom) and a part_type (SMD/TH). 

I understand that this type member was derived originally from the BRD file format, but I have started to implement GenCAD support, and I did not wanted to deal with magic numbers.

martonmiklos:
Okay, I have checked the stylistic issues with clang-formatter, it also changed some other (previosuly untouched by me) code. (I have ran it with clang-format -i -style=file src/openboardview/FileFormats/* way.)

I have separated the mounting_side and part_type properties, as suggested.

piernov:
Thanks, looks fine to me.
I guess the numeric constants could be removed from the `enum class BRDPartMountingSide`?

Yeah, I probably forgot to run `CSTFile.cpp` through `clang-format` before merging the changes…
Also, you could add your `QtCreator project files` to `.gitignore` if you don't want them to be taken into account by `git`. (best is to individually add each files for each commits though)

martonmiklos:
Yes the numeric values has no longer meaning, I will remove them. 

I will add the project files to the .gitignore, the IDE seems to tricked me several times by adding them to git.

martonmiklos:
Squashed.

piernov:
We could have squashed from Github but that's fine anyway.
Seems like there are not many maintainers left here so I'm merging this. Thanks!

MattSturgeon:
LGTM, always in favour of improving readability!

PR #115, opened 2017-10-28
Fix memory leak happening when opening file

Nazg-Gul:
The issue was caused by stringfile() allocating memory for lines pointers, and
this memory was never freed (even though the comment of the function explicitly
mentioned that the result is to be freed).

This commit replaces allocation of bare memory with using std::vector to store
pointers to all lines. This saves us a trouble of manually freeing memory.

Most of the board parsers became shorter, but some of them are still using
bare pointers internally because they need to pass line pointer around and
modify it (i.e. skip lines). It is probably possible to make that code more
modern, but could happen as a separate change.

### DIFF
--- a/src/openboardview/FileFormats/ASCFile.cpp
+++ b/src/openboardview/FileFormats/ASCFile.cpp
@@ -12,9 +12,9 @@
 buf) );
 }*/
 
-void ASCFile::parse_format(char *&p, char *&s, char *&arena, char *&arena_end, char **&lines) {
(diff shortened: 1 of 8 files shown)


Nazg-Gul:
Updated some areas to be more C++-ish (mainly, to avoid bare pointers, which requires some special NULL terminator in the vector). This solved bad memory access in some formats (bad memory access was introduced by the original change).

On a separate but related topic, what abut having a regression tests which will include all supported file formats? So changes in file format parser will be really easy to test.

PR #117, opened 2017-11-09
Fix compilation error on Windows using MSVC 2015/2017

Nazg-Gul:
Several changes in this commit:

- -Werror=format-security was causing compiler to abort
   with uncnown command line argument.

- Make it possible to compile 64bit builds and link against
  64bit SDL (was hardcoded path on Windows).

- MSVC does not define F_OK which was used for access(),
  do ow define in platform comatibility layer.

- No need to do any magic to get case-insensitive file
  path on Windows. Files are insensitive to case anyway, so
  can avoid logic which was failing to compile (no dirent
  on MSVC).

- codecvt linking error is still not fixed in MSVC 2017.


MattSturgeon:
Haven't booted up windows to test this out, but it looks good to me. Merging for now, thanks!

PR #125, opened 2018-04-22
Only link against libdl on Linux

melak:
Pretty sure Linux is the only one needing `libdl`, cf.

https://linux.die.net/man/3/dlopen
https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/dlopen.3.html
https://www.freebsd.org/cgi/man.cgi?query=dlopen&sektion=3
https://man.openbsd.org/dlopen.3
http://netbsd.gw.com/cgi-bin/man-cgi?dlopen+3+NetBSD-current
https://www.dragonflybsd.org/cgi/web-man?command=dlopen&section=3
https://docs.oracle.com/cd/E26502_01/html/E29034/dlopen-3c.html
https://docs.oracle.com/cd/E86824_01/html/E54772/libdl-3lib.html

### DIFF
--- a/src/openboardview/CMakeLists.txt
+++ b/src/openboardview/CMakeLists.txt
@@ -154,7 +154,7 @@ else()
 		${SDL2_LIBRARIES}
 	)
 endif()
-if(NOT WIN32)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 	target_link_libraries(${PROJECT_NAME_LOWER}
 		dl
 	)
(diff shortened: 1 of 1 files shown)


src/openboardview/CMakeLists.txt
```diff
@@ -154,7 +154,7 @@ else()
 		${SDL2_LIBRARIES}
 	)
 endif()
-if(NOT WIN32)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
krytarowski: Please use something like `target_link_libraries(${PROJECT_NAME_LOWER} ${CMAKE_DL_LIBS})`

krytarowski:
The proper macro is: `CMAKE_DL_LIBS`. Haiku requires 2 libraries `be` and `core`, AIX `ldl`, HPUX `dld`. Modern systems like NetBSD have the `dlopen`(3) functionality built into libc.

melak:
That's much cleaner indeed, cheers! Builds fine on Linux and FreeBSD.

MattSturgeon:
Looks good, thanks!

PR #136, opened 2018-07-16
Reload function

TerminalMan:
Added board reload functionality. 'reload currently opened file'
File -> Reload
CTRL+r

+ some minor necessary changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/share/
 /Debug
 /Release
 /release_build/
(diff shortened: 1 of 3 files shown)


inflex:
I can understand this being used in a custom fork of OBV for developments but unsure if it's something that is applicable in the main release, as I understand it you're using it so you can quickly check on your boardview file generation ?

If it's felt that this feature should be added,  I'd say it may be better to modify the existing Load() call such that if it looks for a flag to preserves the state rather than the separate loader.  That way we don't have to remember to duplicate  code for any new file format being added in to the Load call in the future.

MattSturgeon:
I think it's best to avoid uneccessary feature bloat, however if something is _especially_ helpful for development it may be worth adding it behind a build flag - so long as it doesn't get in the way of the main codebase and the added complexity is worth the extra convenience.

PR #142, opened 2018-08-18
Some tweaks to support Lenovo brd format variant (ex. T420)

piernov:
Need more files to check consistency of this variant. Some nets aren't referenced in the file so a lot of pads are left unconnected apparently.


src/openboardview/BoardView.cpp
```diff
@@ -2914,6 +2914,9 @@ inline void BoardView::DrawOutline(ImDrawList *draw) {
 	Point fp;
 
 	auto &outline = m_board->OutlinePoints();
+	if (outline.size() < 1) { // Nothing to draw
+		return;
```
inflex: Good safety pick up, thanks.

inflex:
Looks good to me, thanks.

PR #145, opened 2018-10-13
Fix get_user_dir for Linux when XDG env var is set

zraexy:
If XDG_CONFIG_HOME and XDG_DATA_HOME are set, the current directory is used because they never get added to the path.


piernov:
Good catch, thanks!

PR #153, opened 2019-04-25
Outline failsafe

inflex:
Add a board outline generator as a failsafe for some boardfiles that don't have outlines included in them.


inflex:
Going to sort out the macOS/clang issues and my own OBV fork and come back to this.

PR #154, opened 2019-04-25
Board outline failsafe generator

inflex:
In situations where no board outline has been generated (either lack of data in the board file or parsing incompatibility),  good to have a failsafe outline generated.


inflex:
There's redundancy of the outline generator due to the same sort of code being in several of the board format readers already; however this is a low impact failsafe in the event that something slips through.

Didn't mean to have the pads-both-sides ( b89e755a286c404296de6d835211f602819dc381 ) jump in but being a single line it'll shouldn't cause any real confusion.

PR #155, opened 2019-04-27
Search dialog: add searching by details (part mfg code and pin names)

galkinvv:
The new "Including Part Number" checkbox is added to search dialog.
It enables searching the "msgcode" string attached to each element.

For me it is useful for cases like answering
"Where did they put that P-mosfet in this revision of design?"
(datasheets I'm working with has a lot of msgcode data - part number, Ohm values, even packages).

It's checkbox disabled by default,
since it's UI is a bit less expected then other searches -
the results that are showing below search fields are still names,
not msgcodes, so it is not so immediately obvious
why they are included there.

Also, I'm a not 100% sure of portability
of pointer-to-member-of-template class that I used in Searcher.cpp,
but I do not know actual problems with it.

### DIFF
--- a/src/openboardview/Board.cpp
+++ b/src/openboardview/Board.cpp
@@ -1,0 +1,16 @@
+#include "Board.h"
+
+std::vector<const std::string *> Component::searchableStringDetails() const {
+	return {&mfgcode};
+}
+
+std::vector<const std::string *> Net::searchableStringDetails() const {
+	std::vector<const std::string *> result = {};
+	for (const auto &pin : pins) {
+		result.push_back(&pin->name);
+		if (pin->number != pin->name) {
+			result.push_back(&pin->number);
+		}
+	}
+	return result;
+}
--- a/src/openboardview/Board.h
+++ b/src/openboardview/Board.h
@@ -107,6 +107,8 @@ struct Net : BoardElement {
 	string UniqueId() const {
 		return kBoardNetPrefix + name;
 	}
+
+	std::vector<const std::string *> searchableStringDetails() const;
 };
 
 struct outline_pt {
@@ -215,6 +217,8 @@ struct Component : BoardElement {
 	string UniqueId() const {
 		return kBoardComponentPrefix + name;
 	}
+
+	std::vector<const std::string *> searchableStringDetails() const;
 };
 
 class Board {
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -1772,6 +1772,9 @@ void BoardView::SearchComponent(void) {
 			ImGui::SameLine();
 			search_params_changed |= ImGui::Checkbox("Nets", &m_searchNets);
 
+			ImGui::SameLine();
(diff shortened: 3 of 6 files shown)


inflex:
Thanks for the PR, a useful expansion, though not a lot of files tend to carry the meta data on the parts, so might be worth seeing if the feature can be muted on board files that do not have such information too ( just in case someone tries searching, expecting results and then looks to blame OBV when nothing comes up).

I'd prefer to see the bool to activate the meta search as a public member in the object (or private and set by method);

ie   search.parts_search_meta = true/false; // could use this for the checkbox

That way we can do away with changing the method definition; the other way maybe better, we could go with;

search.setPartsOptionalFields(Component::mfgcode); // keeps us from baking in the fields in the object

I considered a bitmapped parameter to select the fields but would rather keep the searcher as free of specifics as much as possible.

galkinvv:
Thanks for comments. I've made the "where to search" parameter configured via separate field.

Also, after this change I found that it would be more logical to implement two independent checkboxes "Components name" and "Components part number" (instead of additive "including part number"). But I'm not sure, maybe additive variant is better.

P.S. English is not my native, so plural/singular in checkbox texts may be wrong.

Change is force-push (rewritten history instead of new commit,this works fine for gitlab reviews, not so sure here...). Hope that is ok.

galkinvv:
Shall I make some more improvemnets to this PR?
I interpreted previous review quite "wide" and maybe introduceed too much changes - that looked the most readable variant to me. However I undestand that readabilty by mantainer is important.
So I'm open to more reviews )

The search UI window of modified version looks like this:
image

galkinvv:
"Patch status update": I'm using version patched with this search for ~2 years, and mostly it is useful & stable.

However, it has one UI issue (that I document here either for myself in the future or for anyone who may be intereseted in this patch).

If "Components part number" is the only selected search mode - than during typing eements are filtered in the first column below input text field. But upon clickng on one of them (intuitively, to show only that element, like in component or net search mode) - the input field become populated with element name... and it is not found since search by component name is not enabled.

This maybe can be fixed by disallowing searching part number without searching components.This will limit seach flexibility a bit, but will be a quick fix for the UI bug.

galkinvv:
I messed up with this PR (forgot about it and pushed unrelated changes to my master branch...)
Now fixed back. If I mess up it again - the really wanted change is 1 commit here: https://github.com/galkinvv/OpenBoardView/commits/search-by-mfg-part

And it is still WIP due to the UI problem in previous message. Sorry for the noise, it was github misusage from me.

galkinvv:
Now, after several rewritings - I think this PR is eventaully ready fo review.
The first commit is exactly #249 (will be rebased), and the second is adding new search functionality.

Example usage (in scale optimal for screenshot pin highlighting is less distinguashable than in day-to-day usage scale).
Fisrt column demostrate finding elements by MfgCode, second - finding nets by pin name
image

galkinvv:
After #249 was merged, this is rebased and now just contains single commit.

@piernov - this is the last PR I have for openboardview for now; all other custom patches that I was using - already upstreamed.

PR #164, opened 2019-08-07
Extend gitignore

martonmiklos:
Adding more generated/unrelated files to the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,39 @@ imgui.ini
 .idea/
 tags
 
-# Do not track QtCreator imported project files
+# Do not track QtCreator imported or any other project files
 OpenBoardView.config
 OpenBoardView.files
 OpenBoardView.includes
+*.creator
+*.creator.*
+.qmake.stash
+*.pro
+*.pro.*
+
+# Dolphin hidden files marker
+.directory
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -25,7 +25,42 @@ imgui.ini
...
+CPackConfig.cmake
+CPackSourceConfig.cmake
+
+include/version.h
+asset/openboardview.appdata.xml
+asset/openboardview.desktop
+share/appdata/openboardview.appdata.xml
+share/applications/openboardview.desktop
+share/icons/hicolor/scalable/apps/openboardview.svg
+share/mime/packages/openboardview.xml
```
MattSturgeon: Is it worth just ignoring /asset and `/share`?

martonmiklos: The asset folder contains files in git:
https://github.com/OpenBoardView/OpenBoardView/tree/master/asset

But with the share I agree with you. Give me a minute to fix it.

MattSturgeon: If we already have the directory I'd probably prefer having a .gitignore file _in_ the asset dir

martonmiklos: Makes sense, fixed it!

PR #176, opened 2020-05-14
Board background image

piernov:
##  Board background image — please test! (don't merge yet)

OpenBoardView can now show a background image for the board under the rendered boardview.

Branch is based on #175 just because that makes it easier for me.
~~SDL2_image is added as a dependency to be able to load the common image formats. Using pkgconfig to import into CMake right now, may be a problem with Windows build, not tested yet.~~  It's been replaced by stb_image since SDL2_image was causing issues with Windows build regarding static linking.

There are several parameters that can be adjusted in the background image preferences (saved as `.conf` next to the boardview file)

Remember that OpenBoardView is not an image processing application. Meaning that your pictures will most likely have to be pre-processed with a real image processing program before being usable in OBV. Cropping, rotation, image enhancement and this kind of transformations will not be part of OBV. The few settings are only provided to make it easier to line up the image with the boardview, but the image must be cropped and made straight beforehand.

If you do take nice pictures, please share them along with the `.conf` file.

This is what it looks like:

Screenshot_2020-05-14_02-13-03
Screenshot_2020-05-14_02-12-42

### Testing files:
[820-3209.zip](https://github.com/OpenBoardView/OpenBoardView/files/4625323/820-3209.zip)
(using flatbed scanner, that's why it looks bad especially CPU side))
[820-3115.zip](https://github.com/OpenBoardView/OpenBoardView/files/4808087/820-3115.zip)
(using smartphone camera)


inflex:
Nice job on this one, been a feature in want for a while.  Will try test for you in the coming weeks.

piernov:
Windows binary:
[openboardview.zip](https://github.com/OpenBoardView/OpenBoardView/files/4625676/openboardview.zip)

Hopefully it works because SDL2_image was built for dynamic library loading so the additional DLL are needed for now.

MartijnKuipers:
I have given this a spin on OSX and is looks really nice. After some playing around I settled on using no transparency on the background images and changes some pin-colours to be lighter. Basically my changes are:
- no board or parts fill
- use white for selected part
- use white for NC pins
- use blue for GND pins
- use yellow for other pins

Attached a screen shot on a retina display:
image

Some ideas to improve:
- a key to toggle the image (sometimes it gets into the way)
- a bit thicker netweb lines, as they are hard to see (actually pins could be a bit thicker too)

For now, I am going to use it like this. Excellent. Now we just need to get the  board images :-)

piernov:
https://github.com/piernov/OpenBoardView/releases/tag/8.90.3

PR #181, opened 2020-06-22
openboardview: add "unix" X11 geometry flag handling

dithotxgh:
Add the unix/linux specific X11 XParseGeometry() along with the '-g
<geometry>' flag to handle window sizing and positioning.

To get there, added 2 fields to the 'struct global', tweaked the
parse_parameters() to set the new values if provided on the command
line.

Refactored the call to SDL_CreateWindow() to always use the global
values start_x and start_y, which both now default to
SDL_WINDOWPOS_CENTERED.

Added the X11 shared library target only if not on WIN32 to
src/openboardview/CMakeLists.txt.

### DIFF
--- a/src/openboardview/CMakeLists.txt
+++ b/src/openboardview/CMakeLists.txt
@@ -147,6 +147,10 @@ target_link_libraries(${PROJECT_NAME_LOWER}
 	${CMAKE_DL_LIBS}
 )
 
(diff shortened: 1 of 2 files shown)


dithotxgh:
This is my first pull request ever, so I am sure there is something funky about it.

PR #186, opened 2020-07-19
Utilize BRDPartMountingSide enum for the nails side marking

martonmiklos:


martonmiklos:
Hi folks!

Any feedback on this one?

martonmiklos:
@piernov 
Could you please look into it and merge if meets your expectations or feedback if not?
Thanks in advance!

src/openboardview/BRDBoard.cpp
```diff
@@ -61,7 +61,7 @@ BRDBoard::BRDBoard(const BRDFile *const boardFile)
 			net->is_ground = (net->name == "GND");
 			net->number    = brd_nail.probe;
 
-			if (brd_nail.side == 1) {
+			if (brd_nail.side == BRDPartMountingSide::Top) {
 				net->board_side = kBoardSideTop;
 			} else {
```
piernov, line 66: The case for `BRDPartMountingSide::Both` should be added as well for consistency.

piernov (commented):
Maybe it would be a good idea to do the same for `BRDPin::side`?

Issue #172, opened 2020-03-28
feature request: copy part number/model/pin designator


raszpl:
When clicking on a part there is a small annotation, often including part model/value. 
There are no key binds/mouse click options of copying part number (for example 'R100') nor  associated internal component annotation (for example '10K ohm 1/16W') :( Same for part pin designators.
Something as simple as Right clicking on Info Pane component symbol/annotation/pin designator triggering copy to clipboard would be lovely.

piernov:
Implemented by @galkinvv in #200

PR #200, opened 2020-12-13
Copy to clipboard for component and its pins.

galkinvv:
This implements feature request #172
"Copy" small button is added to the first line of a component.
It copies multiline text, containing
reference designator (and description if present) on 1st line,
pin names and associated in a separate line for every pin

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -1235,6 +1235,24 @@ void BoardView::ShowInfoPane(void) {
 
 			ImGui::SameLine();
 			ImGui::Text("%ld Pin(s)", part->pins.size());
+			ImGui::SameLine();
+			{
+				char name_and_id[128];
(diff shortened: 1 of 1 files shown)


galkinvv:
I noticed that ths year the project is active again. Thanks for the 8.0 version!
After looking in issues I found that I'm not the only who wants the copy functionality, so I've implemented it.

The button looks like
copy-ui

While pressing it the above image - the copied text would be:
```
U4 FLASH AT25F512AN-10SU-2.7//ATMEL 512K  SOIC-8
1 ROM_CS*
2 ROM_SO
3 3V3_F
4 GND
5 ROM_SI
6 ROM_SCLK
7 3V3_F
8 3V3_F
```
without the trailing newline

piernov:
Thanks for the contribution. I'm not entirely sure of the usability as-is, but it's still a good addition and it can be improved later if need be.

PS: project isn't very active, just a commit here and there but contributions are always welcome.

PR #211, opened 2021-04-25
Improve ground pins handling

galkinvv:
The first commit improves ground rendering: it is rendered in distinguished color among
the set of pins highlited by belonging to the same component.

The second commit improves detection of ground pins - they weren't detected for some files


galkinvv:
Here is a sample with ground rendering for a selected component. Previousely ground wasn't rendered for a selected component, so after selecting it was hard to quick-distinguish ground pins


(in the github's image scale it is still hard to distinguish ground, however it is possible in actual scale and the problem of distinguishing is the same that for unselected pins. This commit doesn't make it better or worse, it is only improving behviour of selected component pins)

piernov:
Thanks for the contribution.

PR #219, opened 2021-09-20
Add support for Teradyne GenCAD file format

martonmiklos:
This PR add support for Teradyne GenCAD file formats.
I have squashed the work commits to one, but a lot of kudos goes for @galkinvv for his work.


piernov:
Thanks for the submission, a few points I'd like to see addressed if possible:

- This is quite a large PR, there seems to be at least 4 different things going on here:
  1. BRDPartMountingSide use in other formats
  2. BRDFileBase
  3. mpc dependency
  4. Actual GenCAD implemenation
  
  Ideally i. and ii. should've been in a separate PR but this isn't really an issue. However it'd be nice if possible to have separate commits for ii. and iii.

- mpc should be added to the CMake build system as a library in `src/CMakeLists.txt` following the example of other libraries
- I haven't looked into mpc yet but if possible adding additional dependencies should be avoided if it's just for a single format. It is always a nightmare to have everything build properly for the 3 main OS (+Android) and not break with new releases of compilers and libraries. Also quick look at the README says `mpc Only supports ASCII.` which is quite unfortunate, not really an issue but no support for Unicode is a huge drawback in general these days. Anyway if there's no easy way to avoid mpc I'm still willing to add it, hopefully future parsers can also make use of it if it's useful.
- `GENCADFileGrammar.h` should preferably not be included in the repository but generated on the fly when running CMake.
- `GENCADFileBnf.h` (and `GENCADFileGrammar.h` if it is really required) should be called `GenCADFileBnf.h` to follow the same case as `GenCADFile.h`.
- Maybe `generate_grammar_header.sh ` would find a better place in `utilities/`
- Try to follow the code formatting conventions of the rest of the code base especially for the curly bracket placement (I know it's not always consistent but better try to keep it as clean as possible). You can use `clang-format` to reformat the file automatically (double check what it does, sometimes it may do something stupid).

I haven't looked at the code itself or even tested it yet, will do that in the next few days hopefully.

martonmiklos:
Hi @piernov 

Many thanks for your quick response! I will try to addressyour questions and fix the rest. 

- mpc: I have chosen the BNF based approach for implementation because I managed to find official specs for the format and my idea was that creating a parser is easier this way than manually writing parser code. In the end I still feel this somewhat true.
- mpc Only supports ASCII. - true, but the GenCAD is an ASCII only format. 
- GenCAD casing in filenames - FIXED
- code generator moved to the utilities folder

galkinvv:
@martonmiklos 
I'm using this branch and found that it introduced a crashed on unsupported file format (due to accessing `file` pointer being `nullptr`)

You can cherrypick the fix from https://github.com/galkinvv/OpenBoardView/commit/bfd26d5a5f234e967486854cc57b1764ebadf182

piernov:
@martonmiklos @galkinvv would it be possible for you to send an updated pull request with the following changes?
1. Rebase on master with the latest fixes from both of your branches.
2. Split all the changes not strictly related to this file format support into other pull request(s).
3. Add `mpc` to the CMake build system like other libraries
4. Remove auto-generated files from the repo and instead generate them on the fly with a call from CMake.
5. Try to have a more consistant code formatting for example with curly brackets and whitespaces placement (`clang-format` can be used although it needs a manual check afterwards).

Hopefully it can be included in a future 10.0 release.

swiftgeek:
Added some initial changes to make it work with KiCAD's export in fc2b0fcf8bc7ebc84829bc5fe20e48b008a782cf , but so far I only made it pass through some initial parsing errors (It is still compliant to GenCAD 1.4 spec)

galkinvv:
@martonmiklos @piernov I've tried implementing changes 1,2,3 and 5.

@swiftgeek - I'm sorry, but it seems that your work would need to be rebased/cherrypicked since the branch I linked (and you forked) contains some gencad-unrelated commits from my experiments. My fault, I shouldn't link to such "dirty" branch as a reference in public discussion.

You can rebase either on the link below or wait for the full feature merge since the linked variant maybe adapted more during review.

I "described" changes in 4 commits as can be seen there: https://github.com/galkinvv/OpenBoardView/commits/gencad-rebased-on-9-90
* A: refactoring nail side
* B: introducing mpc
* C: the main implementation by @martonmiklos
* D: formatting commit

The main implementaion since last year was adapted to work with version 9.90.
And a small functional change: the VIAs are now parsed as Both-side pins (like most of them physically are). If this looks visually overcomplicated - I can revert to the behaviour when thay are shown only at bottom side. But personally during working with .cad files for the last year - I'd prefer 2-side VIAs display.

I plan to create separate merge requests for A and B and edit this message adding links to them in several minutes.
Comments are welcome!

About "4." - performing auto-generating `GenCADFileBnf.h` at build time - this would be good, but the current converter is too complex and platform-specific. For C++based crossplatform projects I've done such preprocessing via .cmake scripts but this case look more complex than typical, so I'm not sure that this can be rewritten to "cross-platform compile time" without requiring something like python.

galkinvv:
Also a finding that I found during experimenting with gencad files. For huge files their parsing is stil quite slow, so user-visible speed depends o the optimization level.

Еhe `build.sh` in the root of project tree does NOT pass `CMAKE_BUILD_TYPE=Release` to the cmake, so actually in the `release_build` a build **without** optimizarions is created.
I've manually called `cmake -DCMAKE_BUILD_TYPE=Release` in another dir and parsing a huge gencad file become 6seconds instead of 12 seconds (2x improvement).

So for publishing actual releases with gencad enablng optimizations is important, and CMake by default doesn't enable them.

swiftgeek:
I have rebased my changes on top of https://github.com/galkinvv/OpenBoardView/commit/474fc19a84fdef92aee3dea7c8c873c2c3b63288
in https://github.com/swiftgeek/openboardview/tree/kicad-rebased-on-9-90

With https://github.com/swiftgeek/OpenBoardView/commit/be632813546d72004492f5458958d8e41629edb2 I can open 
GenCAD files exported from KiCAD and it shows board outline/components (and string pin names like for BGA), but nets are still missing.

For testing I was using olimex board from https://github.com/OLIMEX/S3-OLinuXino/tree/master/HARDWARE

swiftgeek:
@galkinvv I have prepared a file with board outline that still has issues importing properly
[board-outline.zip](https://github.com/OpenBoardView/OpenBoardView/files/8843636/board-outline.zip)

<details>
  <summary>Screenshots with expected behaviour/current state</summary>
  
 Expected board outline
board-outline-expected

Current OBV render:
board-outline-current
  
</details>

piernov:
@galkinvv 
> * C: the main implementation by @martonmiklos
> * D: formatting commit

These 2 can be squashed together.

> About "4." - performing auto-generating `GenCADFileBnf.h` at build time - this would be good, but the current converter is too complex and platform-specific. For C++based crossplatform projects I've done such preprocessing via .cmake scripts but this case look more complex than typical, so I'm not sure that this can be rewritten to "cross-platform compile time" without requiring something like python.

I haven't looked into it at all and I don't know how any of this works, but note that for now we build either from Linux or macOS, the Windows build being cross-compiled by MinGW. Also, we already depend on Python indirectly for `glad`. Calling a script from CMake would be fine, I simply want to avoid adding automatically generated files to the repo, for which one would have to manually call a script to update anyway.


> Еhe `build.sh` in the root of project tree does NOT pass `CMAKE_BUILD_TYPE=Release` to the cmake, so actually in the `release_build` a build **without** optimizarions is created.
> I've manually called `cmake -DCMAKE_BUILD_TYPE=Release` in another dir and parsing a huge gencad file become 6seconds instead of 12 seconds (2x improvement).

I haven't noticed that, it can probably be made the default in the top CMakeLists.txt, you can also submit a PR for that.

@swiftgeek 
> I have prepared a file with a board outline that still has issues loading properly

There was something similar happening in BVR3 format due to having a segmented outline mode which OBV doesn't know how to handle, no idea if it's related but you can look at the discussion in #224 and the code in "BVR3File.cpp".

swiftgeek:
Got nets working with yet another commit! https://github.com/swiftgeek/OpenBoardView/commit/2d6f59ac3a15d4c860fcd3cb2f794afdbfb66451

There are still some issues where I'm too confused how to fix it, like `GenCADFile::get_padstack_by_name`

martonmiklos:
> @martonmiklos @galkinvv would it be possible for you to send an updated pull request with the following changes?
> 
>     1. Rebase on master with the latest fixes from both of your branches.
> 
>     2. Split all the changes not strictly related to this file format support into other pull request(s).
> 
>     3. Add `mpc` to the CMake build system like other libraries
> 
>     4. Remove auto-generated files from the repo and instead generate them on the fly with a call from CMake.
> 
>     5. Try to have a more consistant code formatting for example with curly brackets and whitespaces placement (`clang-format` can be used although it needs a manual check afterwards).
> 
> 
> Hopefully it can be included in a future 10.0 release.

Hi guys,

I apologize for disappearing from the topic for a while, but it totally went out from my mind that some cleanup needs to be done here before upstreaming. I appericiate the work of @galkinvv for cleaning up my mess ;)

galkinvv:
Most Issues including using python during buildtime grammare generaion adressed in https://github.com/OpenBoardView/OpenBoardView/pull/246

The MPC lib is separate PR https://github.com/OpenBoardView/OpenBoardView/pull/244



The only non-addressed issue left is squashing the initial code and whitespace formatting. I can do it if you are sure it is needed, but I don't think thats a good idea.
The reason in separate commits is simplifying git workflow for anybody who has some local changes above this gencad format PR. If we squash the commits - it would be harder to cherry-pick+merge such changes.
I haven't any changes above it, but for example @swiftgeek had such work (now already rebased over whitespace formatting) and maybe somebody-on-the-internet has some changes also.

PR #220, opened 2021-10-25
a little giant leap

bunker-bunk567:
Hi there.

i worked a couple of weeks on openboardview and am now looking for a way to publish.

included changes:

TCL embedded scripting  inspired by xilinx vivado

pdf cross referencing

heuristic discovery of circuit and board layout / functional groups (using TCL)

drawing resistors as zig-zags and capacitors as plates on the board (very pretty)

### DIFF
--- a/src/openboardview/TCL.h
+++ b/src/openboardview/TCL.h
@@ -41,7 +41,6 @@
 
 #include "BoardView.h"
 #include "platform.h"
-#include "imgui_operators.h"
 #include <regex>
 #include <atomic>
 
--- a/library.tcl
+++ b/library.tcl
@@ -3,6 +3,9 @@ package require Thread
 
 set trace_frame 0
 
+#source "share/openboardview/scripts/find_vrm.tcl"
+#source "share/openboardview/scripts/find_pins.tcl"
+
 proc trace_execution { cmd code result op } {
 	global trace_frame
 	if { $code == 1 } { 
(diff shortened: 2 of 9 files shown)


bunker-bunk567:
it can be made a compile time option.

a program into which GPL code has been added and from which it has been removed is no longer obliged to be licensed under the GPL.

src/openboardview/TCL.h
```diff
@@ -29,9 +29,13 @@
 
 #include "BoardView.h"
 #include <regex>
+#include <atomic>
 
 //#include "TCLDUMMY.h"
 
+#ifdef OBV_USE_POPPLER
+
+#error Enabling pdf rendering with poppler changes the license of openboardview. Uncomment this line to continue with this derivation of openboardview under GPL license
```
MattSturgeon: Shouldn't this say "Comment-out this line" rather than "Uncomment"?

MattSturgeon: I see you later committed a change commenting out this line. If this is how we're notifying users this error prompt should be un-commented when merged.

That said, ideally it'd be better if there was a prompt created by the CMAKE buildscript - something like "type yes to confirm you understand". Ideally the build script would also bundle a copy of the GPL into the build.

bunker-bunk567: i just leave poppler enabled while i am working on the code. since a merge into openboardview is not in the immediate future, i don't care about it at the moment.

don't care about the cmake option. anyone able to build openboardview from source should have no problem following simple instructions. even more true about anyone planning to do a closed source derivative work.

bunker-bunk567: yes. comment it out. for sure. sorry bout that.

PR #223, opened 2021-11-14
Misc patches for Haiku support

mmuman:
These patches makes it run on Haiku.
- use more portable way to count CPU cores in build.sh
- really make GTK optional, we don't have it.
- cache some `tput` results, devices runtime by 3 in a Haiku VM. Even on Linux it devices by 5 user time.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,17 +20,7 @@ EOH
 
 PROJECT="$(color 3 OpenBoardView)"
 if [ -z $THREADS ]; then
-    THREADS=1
-    case "$(uname -s)" in
-      *Darwin*|*BSD*)
-        THREADS=`sysctl -n hw.ncpu`
-        ;;
-      *)
-        if [ -r "/proc/cpuinfo" ]; then
-            THREADS=`cat /proc/cpuinfo | grep processor | wc -l`
-        fi
-        ;;
-    esac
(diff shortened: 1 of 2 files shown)


mmuman:
I just noticed we already had [a different patch around at HaikuPorts](https://github.com/haikuports/haikuports/tree/master/sci-electronics/openboardview/patches), with a native file picker.

Another good option might be [using nativefiledialog instead](https://github.com/mlabbe/nativefiledialog/).

piernov:
These changes are not really specific to Haiku but that's fine.

For proper Haiku support I can merge the HaikuPorts' file picker implementation if it makes any sense. However it would have to be in a separate file (like `haiku.c`) and the other changes about SDL2 should be dropped.

About nativefiledialog, it doesn't seem to bring support beyond what we already have (which doesn't take that many lines of codes and is "good enough") and we try not to add additional external dependencies as much as possible.

mmuman:
> For proper Haiku support I can merge the HaikuPorts' file picker implementation if it makes any sense. However it would have to be in a separate file (like `haiku.c`) and the other changes about SDL2 should be dropped.

Yeah I'll have to see with @threedeyes who did this patch.

> About nativefiledialog, it doesn't seem to bring support beyond what we already have (which doesn't take that many lines of codes and is "good enough") and we try not to add additional external dependencies as much as possible.

Indeed dependency hell is not better.

Issue #224, opened 2021-11-15
BVR Format isnt working


HeyWatchOutDude:
Hi,

is the ".BVR" format not supported by the "OBV" software?

Cant open it.

inflex:
What's the first line in the BVR file?  I'm guessing it'll be a BVR3 version file, as opposed to BVR1 that OBV currently has implemented. 

BVR3 needs to be ported over but there's other items in my queue that I have to sort out first.

HeyWatchOutDude:
Example file: 820-02020_V1.0.bvr

inflex:
Yes, that's a version-3 BVR  file, likewise will be the case with the 820-02016 board and any other new ones that come out.

BVR (v3) is in FlexBV currently but needs to be backported to OBV, or someone else can write their own decoder for OBV ( it's relatively trivial, BVR v3 is explicitly verbose in order to make parsing extremely simple )

HeyWatchOutDude:
Will it be supported in the future?

inflex:
I dare say it will be, yes.  Just a case of finding the time to do it ( I'm busy working on Teboview currently )

HeyWatchOutDude:
Ok thanks for your help so far :)

Tedworthy:
I was trying to access the exact same BV3 file yesterday for the A2338 MacBook Pro, but didn't want to purchase FlexBV just to open a single board. After reading the comments here, then comparing the BVR3 file with a BVR1 file to understand the format, I thought I would try to write a decoder for BVR3.

I'm pleased to say that I was successful and have been able to open this BVR3 file and see the board. All the parts and pins appear to be loaded perfectly, but I have come across an issue with the board outline ('format' in BRDFile). You can see the extent of the issue by checking the screenshot of OBV below and then comparing against a physical PCB photo from iFixit:

<img width="1083" alt="Screenshot 2022-04-18 at 11 02 51" src="https://user-images.githubusercontent.com/944950/163794298-65575d99-6aac-45cc-a9c5-2778beae78d6.png">
<img width="1090" alt="Screenshot 2022-04-18 at 11 07 12" src="https://user-images.githubusercontent.com/944950/163794315-6c1b5dea-5221-4729-ae2c-938946908dd1.png">

[...]

PR #236, opened 2022-04-23
Add support for BVR3 file format

Tedworthy:
This PR adds support for the BVR3 file format. Thanks to @inflex for the help with understanding the OUTLINE_SEGMENTS section.

As I am working from a Mac and using VS Code, I have also added a few lines to the gitignore that are specific to that environment. I'm aware they're not really relevant for this PR but equally I think they'd be useful for other devs and it seemed a bit cumbersome to create a separate PR for it. Feel free to remove those changes if you don't want them.

### DIFF
--- a/src/openboardview/FileFormats/BVR3File.cpp
+++ b/src/openboardview/FileFormats/BVR3File.cpp
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <cstring>
 #include <list>
+#include <algorithm>
 
 int manhattan_distance(const BRDPoint &p1, const BRDPoint &p2) {
 	return abs(p1.x - p2.x) + abs(p1.y - p2.y);
@@ -72,12 +73,13 @@ BVR3File::BVR3File(std::vector<char> &buf) {
 				part.part_type = BRDPartType::SMD;
 			else
 				part.part_type = BRDPartType::ThroughHole;
+		} else if (!strncmp(line, "PART_OUTLINE_RELATIVE ", 22)) {
+			// Value ignored, custom outline for parts not yet supported
 		} else if (!strncmp(line, "PIN_ID ", 7)) {
-			p += 7;
-			pin.snum = READ_STR();
-			pin.part = parts.size() + 1; // pin is for current part, which will not yet have been added to parts vector
+			// Value ignored for time being
 		} else if (!strncmp(line, "PIN_NUMBER ", 11)) {
-			// Ignored, not sure if relevant for BRDPin
+			p += 11;
+			pin.snum = READ_STR();
 		} else if (!strncmp(line, "PIN_NAME ", 9)) {
 			p += 9;
 			pin.name = READ_STR();
@@ -100,8 +102,9 @@ BVR3File::BVR3File(std::vector<char> &buf) {
 		} else if (!strncmp(line, "PIN_COMMENT ", 12)) {
 			// Ignored, not sure if relevant for BRDPin
 		} else if (!strncmp(line, "PIN_OUTLINE_RELATIVE ", 21)) {
-			// Ignored, not sure if relevant for BRDPin
+			// Value ignored, custom outline for pins not yet supported
 		} else if (!strcmp(line, "PIN_END")) {
+			pin.part = parts.size() + 1; // pin is for current part, which will not yet have been added to parts vector
 			pins.push_back(pin);
 			pin = blank_pin;
 		} else if (!strcmp(line, "PART_END")) {
--- a/src/openboardview/FileFormats/BVR3File.cpp
+++ b/src/openboardview/FileFormats/BVR3File.cpp
@@ -111,6 +111,16 @@ BVR3File::BVR3File(std::vector<char> &buf) {
 			part.end_of_pins = pins.size();
 			parts.push_back(part);
 			part = blank_part;
+		} else if (!strncmp(line, "OUTLINE_POINTS ", 15)) {
+			p += 15;
+			while (p[0]) {
+				BRDPoint point;
+				double x = READ_DOUBLE();
+				point.x  = trunc(x);
+				double y = READ_DOUBLE();
+				point.y  = trunc(y);
+				format.push_back(point);
+			}
 		} else if (!strncmp(line, "OUTLINE_SEGMENTED ", 18)) {
 			p += 18;
 			while (p[0]) {
--- a/src/openboardview/FileFormats/BVR3File.cpp
+++ b/src/openboardview/FileFormats/BVR3File.cpp
@@ -1,0 +1,203 @@
+#include "BVR3File.h"
+
+#include "utils.h"
+#include <cmath>
+#include <cctype>
+#include <clocale>
+#include <cstdint>
+#include <cstring>
+
+bool BVR3File::verifyFormat(std::vector<char> &buf) {
+	return find_str_in_buf("BVRAW_FORMAT_3", buf);
+}
+
+BVR3File::BVR3File(std::vector<char> &buf) {
+	auto buffer_size = buf.size();
+
+	char *saved_locale = setlocale(LC_NUMERIC, "C"); // Use '.' as delimiter for strtod
+
+	ENSURE_OR_FAIL(buffer_size > 4, error_msg, return);
+	size_t file_buf_size = 3 * (1 + buffer_size);
+	file_buf             = (char *)calloc(1, file_buf_size);
+	ENSURE_OR_FAIL(file_buf != nullptr, error_msg, return);
+
+	std::copy(buf.begin(), buf.end(), file_buf);
+	file_buf[buffer_size] = 0;
+	// This is for fixing degenerate utf8
+	char *arena     = &file_buf[buffer_size + 1];
+	char *arena_end = file_buf + file_buf_size - 1;
+	*arena_end      = 0;
+
+	int current_block = 0;
+	BRDPart empty_part;
+	BRDPin empty_pin;
+	BRDPart part;
+	BRDPin pin;
+
+	std::vector<char *> lines;
+	stringfile(file_buf, lines);
+
+	fprintf(stderr, "START\n");
+
+	for (char *line : lines) {
+		while (isspace((uint8_t)*line)) line++;
+		if (!line[0]) continue;
+
+		char *p = line;
+		char *s;
+
+		if (!strncmp(line, "PART_NAME ", 10)) {
+			p += 10;
+			part.name = READ_STR();
+			continue;
+		}
+
+		if (!strncmp(line, "PART_SIDE ", 10)) {
+			p += 10;
+			char *side = READ_STR();
+			if (!strcmp(side, "T"))
+				part.mounting_side = BRDPartMountingSide::Top;
+			else
+				part.mounting_side = BRDPartMountingSide::Bottom;
+			continue;
+		}
+
+		if (!strncmp(line, "PART_ORIGIN ", 12)) {
+			p += 12;
+			double origin_x = READ_DOUBLE();
+			part.p1.x = trunc(origin_x);
+			double origin_y = READ_DOUBLE();
+			part.p1.y = trunc(origin_y);
+			continue;
+		}
+
+		if (!strncmp(line, "PART_MOUNT ", 11)) {
+			p += 11;
+			char *mount = READ_STR();
+			if (!strcmp(mount, "SMD"))
+				part.part_type = BRDPartType::SMD;
+			else
+				part.part_type = BRDPartType::ThroughHole;
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_ID ", 7)) {
+			p += 7;
+			pin.snum = READ_STR();
+			pin.part = parts.size() + 1; // Pin is for current part, but will not yet have been added to parts vector
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_NUMBER ", 11)) {
+			// Ignored, not sure if relevant for BRDPin
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_NAME ", 9)) {
+			p += 9;
+			pin.name = READ_STR();
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_SIDE ", 9)) {
+			// Ignored, not sure if relevant for BRDPin
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_ORIGIN ", 11)) {
+			p += 11;
+			double origin_x = READ_DOUBLE();
+			pin.pos.x = trunc(origin_x);
+			double origin_y = READ_DOUBLE();
+			pin.pos.y = trunc(origin_y);
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_RADIUS ", 11)) {
+			p += 11;
+			pin.radius = READ_DOUBLE();
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_NET ", 8)) {
+			p += 8;
+			pin.net = READ_STR();
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_TYPE ", 9)) {
+			// Ignored, not sure if relevant for BRDPin
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_COMMENT ", 12)) {
+			// Ignored, not sure if relevant for BRDPin
+			continue;
+		}
+
+		if (!strncmp(line, "PIN_OUTLINE_RELATIVE ", 21)) {
+			// Ignored, not sure if relevant for BRDPin
+			continue;
+		}
+
+		if (!strcmp(line, "PIN_END")) {
+			pins.push_back(pin);
+			fprintf(stderr, "BRDPin: {\n");
+			fprintf(stderr, "  snum: '%s'\n", pin.snum);
+			fprintf(stderr, "  name: '%s'\n", pin.name);
+			fprintf(stderr, "  pos: (%d, %d)\n", pin.pos.x, pin.pos.y);
+			fprintf(stderr, "  radius: %f\n", pin.radius);
+			fprintf(stderr, "  net: '%s'\n", pin.net);
+			fprintf(stderr, "  part: %u\n", pin.part);
+			fprintf(stderr, "  side: %u\n", pin.side);
+			fprintf(stderr, "}\n");
+			pin = empty_pin;
+			continue;
+		}
+
+		if (!strcmp(line, "PART_END")) {
+			part.end_of_pins = pins.size();
+			parts.push_back(part);
+			fprintf(stderr, "BRDPart: {\n");
+			fprintf(stderr, "  name: '%s'\n", part.name);
+			fprintf(stderr, "  p1: (%d, %d)\n", part.p1.x, part.p1.y);
+			fprintf(stderr, "  p2: (%d, %d)\n", part.p2.x, part.p2.y);
+			fprintf(stderr, "  mounting_side: %d\n", part.mounting_side);
+			fprintf(stderr, "  part_type: %d\n", part.part_type);
+			fprintf(stderr, "  end_of_pins: %u\n", part.end_of_pins);
+			fprintf(stderr, "}\n");
+			part = empty_part;
+			continue;
+		}
+
+		if (!strncmp(line, "OUTLINE_SEGMENTED ", 18)) {
+			p += 18;
+			while (p[0]) {
+				BRDPoint point;
+				double x = READ_DOUBLE();
+				point.x  = trunc(x);
+				double y = READ_DOUBLE();
+				point.y  = trunc(y);
+				//auto existing_point = std::find(format.begin(), format.end(), point);
+				//if (existing_point != format.end())
+				//	format.erase(existing_point);
+				//if (std::find(format.begin(), format.end(), point) == format.end())
+				format.push_back(point);
+			}
+
+			fprintf(stderr, "format:\n");
+			for (BRDPoint format_point : format)
+				fprintf(stderr, "(%d, %d)\n", format_point.x, format_point.y);
+			continue;
+		}
+	}
+
+	num_parts  = parts.size();
+	num_pins   = pins.size();
+	num_format = format.size();
+	num_nails  = nails.size();
+
+	setlocale(LC_NUMERIC, saved_locale); // Restore locale
+
+	valid = true; // TODO
+}
--- a/src/openboardview/FileFormats/BVR3File.cpp
+++ b/src/openboardview/FileFormats/BVR3File.cpp
@@ -85,9 +85,11 @@ BVR3File::BVR3File(std::vector<char> &buf) {
 			p += 9;
 			char *side = READ_STR();
 			if (!strcmp(side, "T"))
-				pin.side = 1;
+				pin.side = BRDPinSide::Top;
 			else if (!strcmp(side, "B"))
-				pin.side = 2;
+				pin.side = BRDPinSide::Bottom;
+			else if (!strcmp(side, "O"))
+				pin.side = BRDPinSide::Both;
 		} else if (!strncmp(line, "PIN_ORIGIN ", 11)) {
 			p += 11;
 			double origin_x = READ_DOUBLE();
--- a/src/openboardview/FileFormats/BVR3File.cpp
+++ b/src/openboardview/FileFormats/BVR3File.cpp
@@ -58,14 +58,12 @@ BVR3File::BVR3File(std::vector<char> &buf) {
 			char *side = READ_STR();
 			if (!strcmp(side, "T"))
 				part.mounting_side = BRDPartMountingSide::Top;
-			else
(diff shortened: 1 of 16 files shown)


src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+						match_found = true;
+						break;
+					}
+					it++;
+				}
+				if (match_found)
+					continue;
+
+				// Exact match not found, so pick nearest segment instead
+				it = std::min_element(
```
piernov, line 163: Missing `#include <algorithm>`

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+			p += 10;
+			char *side = READ_STR();
+			if (!strcmp(side, "T"))
+				part.mounting_side = BRDPartMountingSide::Top;
+			else
+				part.mounting_side = BRDPartMountingSide::Bottom;
+		} else if (!strncmp(line, "PART_ORIGIN ", 12)) {
+			p += 12;
+			double origin_x = READ_DOUBLE();
+			part.p1.x = trunc(origin_x);
```
piernov, line 65: Note that `BRDPart::p1` and `BRDPart::p2` are only used to draw a rectangle for parts that have no pins, and for that purpose, both points have to be defined.
 I don't know what `PART_ORIGIN` is supposed to mean since it's almost always 0 except in some files with "orphan" probe points (or vias) where it is set to the position of the pin.
For now maybe better to leave it out.

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+			if (!strcmp(mount, "SMD"))
+				part.part_type = BRDPartType::SMD;
+			else
+				part.part_type = BRDPartType::ThroughHole;
+		} else if (!strncmp(line, "PIN_ID ", 7)) {
+			p += 7;
+			pin.snum = READ_STR();
+			pin.part = parts.size() + 1; // pin is for current part, which will not yet have been added to parts vector
+		} else if (!strncmp(line, "PIN_NUMBER ", 11)) {
+			// Ignored, not sure if relevant for BRDPin
```
piernov, line 80: Should be assigned to `BRDPin::snum` since it is what will be used to find and highlight pin number 1.

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+		} else if (!strncmp(line, "PART_MOUNT ", 11)) {
+			p += 11;
+			char *mount = READ_STR();
+			if (!strcmp(mount, "SMD"))
+				part.part_type = BRDPartType::SMD;
+			else
+				part.part_type = BRDPartType::ThroughHole;
+		} else if (!strncmp(line, "PIN_ID ", 7)) {
+			p += 7;
+			pin.snum = READ_STR();
```
piernov, line 77: Not sure what `PIN_ID` is supposed to be, doesn't really sound useful. `BRDPin::snum` should contain `PIN_NUMBER` instead.

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+			pin.part = parts.size() + 1; // pin is for current part, which will not yet have been added to parts vector
+		} else if (!strncmp(line, "PIN_NUMBER ", 11)) {
+			// Ignored, not sure if relevant for BRDPin
+		} else if (!strncmp(line, "PIN_NAME ", 9)) {
+			p += 9;
+			pin.name = READ_STR();
+		} else if (!strncmp(line, "PIN_SIDE ", 9)) {
+			// Ignored, not sure if relevant for BRDPin
```
piernov, line 85: I don't think `BRDPin::side` is used right now. Could be used in case a part has pins on different sides (or both).

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+			double origin_y = READ_DOUBLE();
+			pin.pos.y = trunc(origin_y);
+		} else if (!strncmp(line, "PIN_RADIUS ", 11)) {
+			p += 11;
+			pin.radius = READ_DOUBLE();
+		} else if (!strncmp(line, "PIN_NET ", 8)) {
+			p += 8;
+			pin.net = READ_STR();
+		} else if (!strncmp(line, "PIN_TYPE ", 9)) {
+			// Ignored, not sure if relevant for BRDPin
```
piernov, line 99: Not sure what it is, good old magic numbers… Seem to be related to net type in some files, or always 2 in others. Probably not really useful.

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+		} else if (!strncmp(line, "PIN_COMMENT ", 12)) {
+			// Ignored, not sure if relevant for BRDPin
+		} else if (!strncmp(line, "PIN_OUTLINE_RELATIVE ", 21)) {
+			// Ignored, not sure if relevant for BRDPin
```
piernov, line 103: Most likely custom outline for each pin which we don't support. There is also a `PART_OUTLINE_RELATIVE` not parsed here. Not useful until custom polygons for parts/pins are implemented.

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+		} else if (!strcmp(line, "PIN_END")) {
+			pins.push_back(pin);
+			pin = blank_pin;
+		} else if (!strcmp(line, "PART_END")) {
+			part.end_of_pins = pins.size();
+			parts.push_back(part);
+			part = blank_part;
+		} else if (!strncmp(line, "OUTLINE_SEGMENTED ", 18)) {
```
piernov, line 111: Some files use `OUTLINE_POINTS` instead which is not handled here.

piernov (commented):
Thanks for the submission, I left some comments to hopefully improve the handling of some files.

There is another issue on M1 boards. The M1 CPU (on 820-02016 and 820-02020) now has over 2000 balls. The code that generates parts outlines in `BoardView::DrawParts` caps at 2000, it should still fallback to min/max coordinates of pins but it doesn't seem to work properly and generates a buggy outline.
Easy fix is to bump to 3000, but the code better be rewritten one day…

Tedworthy:
Thanks for the comments and checking the PR so fast, much appreciated 😊

I'm a little rusty with C++ and not familiar with the BVR3 format and how it translates to your BRDFile classes (hence the comments), so your feedback is very helpful. I'll take a look and make some adjustments, and may also ask for more help on the format specifics in the linked issue.

Could you let me know which test files you've been using? Board names are fine if you can't share the files directly, I could just do with more test cases as I've only been checking on a single board so far. I will check the board you referenced here though.

I'll check DrawParts to see if I can do anything but may have to just increase the max number to 3000 for now, as you hinted. Always a judgement to make between quick fixes and rewrites 😉

src/openboardview/FileFormats/BVR3File.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+		if (!strncmp(line, "PART_NAME ", 10)) {
+			p += 10;
+			part.name = READ_STR();
+		} else if (!strncmp(line, "PART_SIDE ", 10)) {
+			p += 10;
+			char *side = READ_STR();
+			if (!strcmp(side, "T"))
+				part.mounting_side = BRDPartMountingSide::Top;
+			else
```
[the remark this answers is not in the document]
Tedworthy: Have asked what the possible values are for this in #224, will modify when we know more

piernov:
The ones I looked at were mainly 820-01987 and 820-02020, but the ones I know of that are in BVR3 format are:
- 820-01124
- 820-01949 (original `.brd` also available)
- 820-01987 (original `.brd` also available)
- 820-01992
- 820-02005
- 820-02016
- 820-02020

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Have asked what the possible values are for this in #224. I agree though that it's probably better to leave it out unless the meaning highlights something we don't know.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: I've looked back at this and I think the reason I didn't populate that was because I didn't know how T and B values from the file translate to BRDPin::side, which is a uint. If you can advise, I'll modify to suit.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Have asked what the possible values are for this in #224, may modify if it turns out the meaning is relevant.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Thanks, will check other files and implement.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Thanks, I added PART_OUTLINE_RELATIVE parsing for potential future use but have left comments on each explaining they're currently ignored due to no support.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Thanks, modified. The value of PIN_ID is now ignored and BRDPin::snum is now populated from PIN_NUMBER.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Thanks, done.

src/openboardview/FileFormats/BVR3File.cpp
piernov: `BRDPin::side` is not really used for anything right now, only assigned in `BRD2File.cpp` with 1 → Top, 2 → Bottom.
In `BRDBoard.cpp` for `BRDNail::side`, 1 → Top, everything else → Bottom.
In `Board.h`, `EBoardSide` for `BoardElement`, 0 → Top, 1 → Bottom, 2 → Both.
It's not consistent in any case. There should be only one `enum` (`enum class` may I say) for everything.
Not useful for now if there are no files that contain a part with pins on different sides.

piernov:
There are differences in the probe points/vias shown between the `.brd` and `.bvr` versions of the 820-01987, I don't know if there are differences in the files themsleves though, but there are differences compared to FlexBV as well.
Screenshot_2022-04-24_13-31-39
Screenshot_2022-04-24_13-31-28
Screenshot_2022-04-24_13-31-18
Screenshot_2022-04-24_13-31-07

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Implemented, looks good to me now on the files I've tried that use OUTLINE_POINTS but would appreciate you checking to confirm.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Thanks for that info. I've modified the code to treat things the same way as BRD2File, but at some stage it will want replacing with enum values that are consistent across the codebase.

Tedworthy:
First of all, regarding the buggy outlines for the M1 CPU parts, I have bumped up the size to 3000 which seems to fix the issue.

Regarding the differences on 820-01987, I think there may be a problem with the BRD version. If you take a look at the screenshot below, you can see that there are loads of test points that do not seem to appear on the BVR3 version (and I can't see them on your FlexBV screenshots either). The weird thing is that they are all GND, and they're peppered all over the board, which to my mind seems wrong. What would be the point of having hundreds of GND test points? Only thing I can think is that they're vias which are showing as test points. I'm inclined not to treat this as a BVR3 format bug, unless you can see any other discrepancies between OBV and FlexBV that need investigating?
Screenshot 2022-04-24 at 13 38 57

I have noticed another issue that is possibly related to the above. It's not a BVR3 format issue either as the same thing appears on the BRD version, but if you check 820-01949 and look for the test point connected to WLAN_CONTEXT_A, it appears right in the middle of U3730. See the screenshot below:
Screenshot 2022-04-24 at 13 49 57
I've checked the BVR file and the relevant test point is defined as such:
```
PART_NAME ...-16133
   PART_SIDE T
   PART_ORIGIN 3734.000 1757.000

   PIN_ID TP-16133
      PIN_NUMBER 1
      PIN_NAME 1
      PIN_SIDE T
      PIN_ORIGIN 3734.000 1757.000
      PIN_RADIUS 10.000
      PIN_NET WLAN_CONTEXT_A
      PIN_TYPE 5
   PIN_END

PART_END
```
The problem here is that this pin should be on the top side of the board, but the screenshot above is of the bottom side. I think this is happening because OBV treats the parts with names starting "..." as dummy parts containing test points and groups them all together under the same part, which is perfectly logical, but as a result the single dummy part has only one side, which causes any test points defined on the other side (this case) to be drawn on the wrong side. Not sure what the best fix is for this, and it's probably sensible to raise a separate issue given that it's unrelated to this PR. What do you think?

piernov:
This is the kind of differences I am seeing between OBV and FlexBV:
Screenshot_2022-04-24_15-36-39
Screenshot_2022-04-24_15-36-13

These test points indeed show up on the wrong side (compared to FlexBV).
As for the `.brd` file, I was just pointing out that there were a lot of differences to the `.bvr` file in OBV, but the `.brd` look very similar between OBV and FlexBV, so there may still be a difference in how they are parsed.
However if it's not a problem specific to BVR3 this can indeed be fixed later.

"Test points" could indeed show up under parts and they would probably vias in that case (if they are on the correct side of course), we make no difference between test points and vias (when they exist) because the file formats don't provide this information anyway.
However in the case of "WLAN_CONTEXT_A" it seems like it shows up on opposite side compared to FlexBV.

There were some changes in b89e755 and 77d34b7 that could be related.
This may be part of the problem: https://github.com/OpenBoardView/OpenBoardView/blob/0b3ccae55086030fa3c61755d10e4a3fbfd681b0/src/openboardview/BRDBoard.cpp#L167

Tedworthy:
I see, thanks for the clarification. I don't have a copy of FlexBV so the screenshots are really useful.

I've done some investigating and as far as I can tell there are a couple of issues. The first one is the problem I highlighted where the pin side of the test point is not always respected, and the other I think is the issue you picked up with `PART_SIDE O` / `PIN_SIDE O`.

On the first issue, the line of code you highlighted could be part of the problem. Taking the `board_side` value from the net makes total sense if you have no further info for the pin, but if `brd_pin.side` is set specifically to top/bottom then I think it ought to take precedence. I tried replacing that single line of code with the following:
```
if (brd_pin.side == 1) {
    pin->board_side = EBoardSide::kBoardSideTop;
} else if (brd_pin.side == 2) {
    pin->board_side = EBoardSide::kBoardSideBottom;
} else {
    pin->board_side = pin->net->board_side;
}
```
Obviously this does take us back to the previous point we discussed where there is no consistent interpretation of the pin side uint values, but the logic of the above is that the pin side is initialised to 0, so the first two conditions wouldn't get hit unless the code in the file format parsing specifically set the pin side to something. I've tried this code on 820-01949 and 820-01987, and it does appear to make the test points generally render on the correct side. See the following screenshots:
Screenshot 2022-04-24 at 23 39 01
Screenshot 2022-04-24 at 23 38 14
The issue I raised with WLAN_CONTEXT_A seems to be solved, and so does the one you raised with CHGR_RST_IN.

It still doesn't look quite right though, for example there are now other test points in the middle of U3730 on 820-01949, and there's still a test point that looks wrong at the top left of R7020 on 820-01987. I checked the BVR3 file data for that test point CHGR_CSI_R_P and it is using `PART_SIDE O` and `PIN_SIDE O`.

I haven't got any clarification yet on what that value means, but looking at this example I think this test point should be on the other side of the board only (it's showing on bottom, should be on top). It's therefore unlikely that it means both sides, but surely it can't mean top or bottom because I've seen other examples of values "T" and "B" for that purpose. I can't think of any other possibilities... Weird! Guess we'll have to wait and see.

What do you think of the above fix? Can you foresee any issues with it, assuming the pin side values are made into a proper enum class like you suggested?

src/openboardview/FileFormats/BVR3File.cpp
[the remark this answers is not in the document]
Tedworthy: Turns out this does mean "both". I have updated the code to suit.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: This is apparently an origin point used when relative points for pins are defined. In any case there is no support for this currently so the origin point is also not useful at present. I have removed this code and added a comment to explain.

src/openboardview/FileFormats/BVR3File.cpp
Tedworthy: Apparently this is really just something in the files for future proofing and whilst it seems it is being set to different values in certain cases, it isn't used much at the moment. I agree it's not useful right now so have updated the comment to reflect this.

piernov:
Thanks for the fixes.
In this case I think it's the file format decoder itself that should handle the side for the pin or test point, that way if the file format provides this information it can be handled accordingly, otherwise the decoder can choose to use the side of the part or net.
Since it depends on the file format, `BRDBoard.cpp` or `BoardView.cpp` should not be making that decision.
It also avoids polluting code outside the file format decoders with magic numbers. (as long as we switch to using a universal `enum class` for the side)
Downside is that it requires adjusting all the other file format decoders.

There's also something weird going on with T2235 in 820-02020:
Screenshot_2022-04-30_14-25-50
Screenshot_2022-04-30_14-25-40

Tedworthy:
@piernov Thanks for checking this, I totally agree about the decoders handling the sides of the pins. I will convert the pin side to an `enum class` and update all the decoders so that those that set pin side do so using consistent enum values.

However I do think we will still need some minor changes to `BRDBoard.cpp`, certainly for test points. Currently the code in this PR sets the pin sides in the decoder as per the BVR3 file, but the sides of test points are not respected when rendered. This seems to be because of how the test points are handled in `BRDBoard.cpp`, at present it checks for the part name being prefixed with `"..."`, and if so it combines all those test points into a single component (the 'dummy' component) and doesn't appear to set `Pin::board_side`. This means the specific sides of the individual test points aren't correctly interpreted so test points sometimes appear on the wrong sides of the board.

Assuming we convert `BRDPin::side` to an `enum class` and set the value consistently, surely the code in `BRDBoard.cpp` should then set `Pin::board_side` directly from `BRDPin::side`, assuming it has been set specifically? The only case where this shouldn't be set directly would be if the file format doesn't support it, in which case the decoder would not have set `BRDPin::side`. Do you agree? I presume this would also mean we'd need to allow for an 'unset' or 'same as part' enum value for the pin side, which would be the default?

Regarding that issue on 820-02020, I actually spotted that myself when testing, however on inspection of the file it's actually an error within the file itself. This is what it looks like for that particular test point and the one immediately before it:
```
PART_NAME T2234
   PART_SIDE B
   PART_ORIGIN 0.000 0.000
   PART_MOUNT SMD

   PIN_ID T2234-TP
      PIN_NUMBER 1
      PIN_NAME TP
      PIN_SIDE B
      PIN_ORIGIN 5463.000 453.000
      PIN_RADIUS 10.000
      PIN_NET NCC1767
      PIN_TYPE 2
      PIN_COMMENT
   PIN_END


PART_NAME T2235
   PART_SIDE B
   PART_ORIGIN 0.000 0.000
   PART_MOUNT SMD

   PIN_ID T2235-TP
      PIN_NUMBER 1
      PIN_NAME TP
      PIN_SIDE B
      PIN_ORIGIN 5247.000 62.000
      PIN_RADIUS 10.000
      PIN_NET NCC2090
      PIN_TYPE 2
      PIN_COMMENT
   PIN_END

PART_END
```
As you can see, there is no `PART_END` tag to complete the T2234 part, it just continues on defining part T2235. Strictly speaking I would consider this an error in the file, but it would of course be possible to detect the fact that the current part hasn't been ended when starting a new part, and work around the error. Considering that FlexBV seems to be doing this, it might be best to do the same for consistency. The same could be done for pins, so that the current pin is always ended if a new pin is started. Do you agree this is the best approach?

piernov:
What I meant is that I think we should rely _only_ on the pin/test point side to draw the pins/test points, so _every_ file format decoder must define them, even if the file format does not provide this information directly (and instead copy the part side if available or net side).
Currently, both `BRDBoard.cpp` and `BoardView.cpp` do not use this information properly, so both need to be modified to use the pin side to populate and draw pins. And they should not try to default to using part/net side since it's not their job to handle file format specific behaviour.
Meaning that there is no need for an "undefined" constant.

About 820-02020, no need to try to workaround malformed files, the file can be fixed an reuploaded. FlexBV doesn't try to draw a part outline for test point parts so that may be the only reason why it doesn't look bad.

Tedworthy:
Thanks for clarifying, that approach makes good sense. I have implemented this now on the PR, hopefully I have understood everything correctly and the latest commit matches what you were expecting.

I can certainly say that things appear to be working well on the test files I have opened. I've done some comparisons between the BRD and BVR versions of a couple of boards and there are now many fewer differences to my eyes. If you could also perform some checks and review the code again that would be much appreciated.

I should also say that whilst I have carefully checked the logic of the changes to the file format decoders, I don't have any test files for the majority of them so I can't verify if the code changes have broken anything. Do you know of any boards that use other file formats?

Fingers crossed we're close to having this PR merged so people can start opening BVR3 files. 🙂

piernov:
I think it's ok, there's still a difference compared to how FlexBV render vias (I think, parts with name starting with `...` and single pin on both sides) since on FlexBV nothing shows up under `U3730` (on the same side) of 820-01949 for example as you pointed out before. This is also how the `.brd` is rendered on OBV. It only shows up when the net is selected. For example selecting pin 15 of `U3730` reveals `TP[1]AP_PCIE_WAKE_L`. It's not that much of a problem but just looks weird. It's possible that FlexBV ignores these parts altogether since there's already a test point rendered separately and there's no bounding box showing up for the part, not sure.

Also, the "nail" side for other file formats is still handled weirdly and with a magic number. It's not related to this PR though.

I'm ok with merging as-is if you do not plan to look into the above points in this PR.

Tedworthy:
@piernov Yeah there is still a difference there, so FlexBV certainly handles this differently in its BVR3 decoding. Here are some snippets from the BVR file relating to `AP_PCIE_WAKE_L` (excluding pin 15 on `U3730`)
```
PART_NAME TP3703
   PART_SIDE T
   PART_ORIGIN 0.000 0.000
   PART_MOUNT SMD

   PIN_ID TP3703-1
      PIN_NUMBER 1
      PIN_NAME 1
      PIN_SIDE T
      PIN_ORIGIN 3883.000 1802.000
      PIN_RADIUS 10.000
      PIN_NET AP_PCIE_WAKE_L
      PIN_TYPE 5
      PIN_COMMENT 
   PIN_END

PART_END
```
and
```
PART_NAME ...-16586
   PART_SIDE O
   PART_ORIGIN 3883.000 1802.000

   PIN_ID TP-16586
      PIN_NUMBER 1
      PIN_NAME 1
      PIN_SIDE O
      PIN_ORIGIN 3883.000 1802.000
      PIN_RADIUS 10.000
      PIN_NET AP_PCIE_WAKE_L
      PIN_TYPE 5
   PIN_END

PART_END
```
You could be right with those suggestions, it's quite possible that FlexBV is noticing that the pin defined on both sides has an identical position to the test point and is therefore choosing to render the test point only.

Another thought I had was that maybe any pin that is defined as a test point _and_ is located on both sides is automatically treated as a via, and is therefore not rendered? I quickly tested this by modifying `BoardView::DrawPins`, inside the main pin loop I added the line `if (pin->type == Pin::kPinTypeTestPad && pin->board_side == kBoardSideBoth) continue;`. Doing this does seem to make the BRD file and BVR3 file match up exactly in and around `U3730`.

Aside from the location of the above code, do you think this is in general a good idea? It seems like it could work, but my knowledge of this stuff is very limited and I'm aware there could be complications or some other good reason as to why it wouldn't work.

Let me know what you think. I'm perfectly happy for the PR to be merged whenever, but equally it would be nice to make the behaviour match FlexBV as much as we can.

piernov:
Vias are a useful bit of information so I don't think they should be forcefully hidden. Even if it looks a bit weird I think they should still be rendered. They could be rendered differently if we actually knew they were vias (instead of test points) but file formats typically do not provide this information. That's assuming what we identified is indeed via and not something else, I don't have any of these boards to confirm, so for now we will leave it as it is.

Let's merge as-is, further adjustments can be done later if required.

Thanks again for all this work.

I guess it's time to release a new stable version of OBV, although I haven't had a lot of feedback about the stability of the beta versions.

PR #243, opened 2022-06-06
Introduce BRDFileBase and cleanup nail handling

galkinvv:
Preparations before introducing gencad format:
* Separate specific BRDFile implementation from common base class BRDFileBase.
* Utilize BRDPartMountingSide for the nails side marking
* Factor out Nails-as-pins conversion to a method of BRDFileBase from specific BRD2 implementation.

### DIFF
--- a/src/openboardview/BoardView.h
+++ b/src/openboardview/BoardView.h
@@ -123,7 +123,7 @@ enum DrawChannel {
 enum FlipModes { flipModeVP = 0, flipModeMP = 1, NUM_FLIP_MODES };
 
 struct BoardView {
-	BRDFile *m_file;
+	BRDFileBase *m_file;
 	Board *m_board;
 	BackgroundImage backgroundImage{m_current_side};
 
@@ -326,7 +326,7 @@ struct BoardView {
 	void DrawParts(ImDrawList *draw);
 	void DrawBoard();
 	void DrawNetWeb(ImDrawList *draw);
-	void LoadBoard(BRDFile *file);
+	void LoadBoard(BRDFileBase *file);
 	int LoadFile(const filesystem::path &filepath);
 	ImVec2 CoordToScreen(float x, float y, float w = 1.0f);
 	ImVec2 ScreenToCoord(float x, float y, float w = 1.0f);
--- a/src/openboardview/FileFormats/BRD2File.cpp
+++ b/src/openboardview/FileFormats/BRD2File.cpp
@@ -152,7 +152,13 @@ BRD2File::BRD2File(std::vector<char> &buf) {
 					std::cerr << "Missing net id: " << netid << std::endl;
 				}
 
-				nail.side = READ_UINT();
+				bool nail_is_top = READ_UINT() == 1;
+				if (nail_is_top) {
+					nail.side = BRDPartMountingSide::Top;
+				} else {
+					nail.side = BRDPartMountingSide::Bottom;
+					nail.pos.y = max.y - nail.pos.y;
+				}
 				nails.push_back(nail);
 
 			} break;
@@ -218,21 +224,7 @@ BRD2File::BRD2File(std::vector<char> &buf) {
 		parts.push_back(part);
 	}
 
-	for (auto &nail : nails) {
-		BRDPin pin;
-		pin.pos = nail.pos;
-		if (nail.side == 1) {
-			pin.part = parts.size();
-			pin.side = BRDPinSide::Top;
-		} else {
-			pin.pos.y = max.y - pin.pos.y;
-			pin.part  = parts.size() - 1;
-			pin.side = BRDPinSide::Bottom;
-		}
-		pin.probe = nail.probe;
-		pin.net   = nail.net;
-		pins.push_back(pin);
-	}
+	AddNailsAsPins();
 
 	valid = current_block != 0;
 }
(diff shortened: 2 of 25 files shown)


galkinvv:
Note: maybe merging this would make https://github.com/OpenBoardView/OpenBoardView/pull/186 obsolete since this PA is actually "extension" of the linked @martonmiklos PR with some more additions for GenCAD.

src/openboardview/BoardView.h
```diff
@@ -305,6 +305,7 @@ struct BoardView {
 	bool m_showColorPreferences;
 	bool m_firstFrame = true;
 	bool m_lastFileOpenWasInvalid;
+	std::string m_lastFileLoadError;
```
piernov, line 308: Orphan attribute

src/openboardview/FileFormats/BRD2File.cpp
```diff
@@ -218,21 +218,7 @@ BRD2File::BRD2File(std::vector<char> &buf) {
...
-		} else {
-			pin.pos.y = max.y - pin.pos.y;
-			pin.part  = parts.size() - 1;
-			pin.side = BRDPinSide::Bottom;
-		}
-		pin.probe = nail.probe;
-		pin.net   = nail.net;
-		pins.push_back(pin);
-	}
+	AddNailsAsPins(max);
```
piernov, line 221: I don't even remember why this piece of code was there. Is it actually useful to be extracted outside of this format?

src/openboardview/FileFormats/BRDFile.cpp
```diff
@@ -77,6 +77,27 @@ char *fix_to_utf8(char *s, char **arena, char *arena_end) {
...
+		BRDPin pin;
+		pin.pos = nail.pos;
+		if (nail.side == BRDPartMountingSide::Both) {
+			pin.part = parts.size();
+			pin.side = BRDPinSide::Both;
+		} else if (nail.side == BRDPartMountingSide::Top) {
+			pin.part = parts.size();
+			pin.side = BRDPinSide::Top;
+		} else {
+			pin.pos.y = max_point.y - pin.pos.y;
```
piernov, line 91: Also I think inverting y coordinate for bottom is specific to BRD2.

src/openboardview/FileFormats/BRDFile.h
```diff
@@ -86,13 +87,20 @@ class BRDFile {
...
-
 	bool valid = false;
 	std::string error_msg = "";
 
+	void AddNailsAsPins(const BRDPoint& max_point);
+	BRDFileBase() {}
+	~BRDFileBase() {}
+};
+
+class BRDFile : public BRDFileBase {
```
piernov, line 98: Would it be possible to rename the current file to "BRDFileBase.h" (containing the parent `BRDFileBase` class, and the `READ_*` macros and shared `BRD*` structs for now) and extract the `BRDFile` class to a new "BRDFile.h" file?
`stringfile()` and `fix_to_utf8()` should also become part of a new "BRDFileBase.cpp" file, even though I'm not sure these functions are all that useful.

src/openboardview/FileFormats/BRD2File.cpp
galkinvv: Yes, it is used for vias in gencad format - see https://github.com/galkinvv/OpenBoardView/blob/474fc19a84fdef92aee3dea7c8c873c2c3b63288/src/openboardview/FileFormats/GenCADFile.cpp#L95

So extracted to base calss to avoid duplication.

src/openboardview/FileFormats/BRDFile.cpp
galkinvv: Yes, thanks. This was unnoticed during gencad development and causing some problems there.
I moved the inverting from AddNailsAsPins to the process of BRD2 reading.
Tested on a BRD2 file with two-side nails that the result is unchanged.

src/openboardview/FileFormats/BRDFile.h
galkinvv: Base class separated into its own file, includes also separated according to their real usage.

galkinvv:
Note the 2nd version of diff adds separate commit fixing pointer double-free-call that seemed to be present for several years.

My version of the fix is simplest, since really I don't no the overall project style plan (towards unique_ptr, towards helper functions or something other).

Also the default-initialization was added to all trivial fields of BRDPoint/BRDPart/BRDPin/BRDNail structures in newly created BrdFileBase, to be more sure that parsing code paths with lots of ifs would never lead to random behaviour due to uninitialized memory.

piernov:
> Note the 2nd version of diff adds separate commit fixing pointer double-free-call that seemed to be present for several years.

Actually the double-free didn't happen because destructor is called through a pointer to base class, and base class' destructor is non-virtual. This is obviously wrong, base class should have virtual destructor and the `free()` should happen in either the base class or all the child classes.

Also, the other file format classes should inherit directly from `BRDFileBase` instead of `BRDFile`, so the `file_buf` attribute will have to be owned either by `BRDFileBase` (in this case it should be freed in `BRDFileBase`'s virtual destructor if non-null), or by all of the child classes.

> My version of the fix is simplest, since really I don't no the overall project style plan (towards unique_ptr, towards helper functions or something other).

Basically: it's a complete mess. In my opinion, static (or stack) allocation or STL's containers should always be preferred, as well as references, and when needed smart pointers. Of course there are situations where you have to use a raw pointers but it's not one of them. Lots of stuff would benefit from a refactor but it's a lot of work, so we just do the best we can when we add new code.

PR #245, opened 2022-06-07
Build in optimized release mode when no CMAKE_BUILD_TYPE type is specified

galkinvv:
Optimized released binaries provides noticable speedup while working with huge boards.


piernov:
Thanks for the contribution.

PR #247, opened 2022-06-14
Make GenCAD support more compliant with the spec

swiftgeek:
Changes in this pull allow for files exported from KiCAD to be opened in OBV, but no change is KiCAD specific and there shouldn't be any regression.

Remaining issues with KiCAD exported files:
* Quotes not stripped from device description - somewhat fixed with https://github.com/swiftgeek/OpenBoardView/commit/47a484f08421619fe9d7bd7f6a0ab2affd3fc416, but not reviewed by anyone yet.
    * not sure if `string_to_end` is a good idea, especially when it can have quoted field instead
* ~~There are still some issues where I'm too confused how to fix it, like `GenCADFile::get_padstack_by_name` `GenCADFile::get_pad_by_name`, `GenCADFile::get_padstack_radius` (They all won't work with quoted strings, as allowed per spec, if I have read it correctly)~~. Implemented following @galkinvv's suggestion
* some ~graphical elements appear as text/components in OBV (using example project: Logo-OSHW, Sign-RecycleBin, Sign-RoHS, Sign-PbFree, etc)
* ~~THT pads only appear on one side~~ #252

For testing I was using olimex board from https://github.com/OLIMEX/S3-OLinuXino/tree/master/HARDWARE

GenCAD file for reference: [S3-OLinuXino_Rev_A.zip](https://github.com/OpenBoardView/OpenBoardView/files/8972216/S3-OLinuXino_Rev_A.zip)


galkinvv:
I've tested the suggested changes with several .cad files that I use - all opens fine.
Also breifly reviewed the code - I don't see any problems.

So, this PR looks good for me!

About 3 methods mentioned:
- `GenCADFile::get_padstack_by_name` and `GenCADFile::get_pad_by_name` utilize the `GenCADFile::has_text_content` method. This method was introduced only as an extended string comparison that checks for string equality with graceful checking the first `mpc_ast_t*` argument for nullptr. No any "unobvious idea" was put in that method, just string comparison wrapper. So, if you change `pad_name` getting to use `get_nonquoted_or_quoted_string_child` - you can just swap `has_text_content` to "nullptr check + string comparison".
- The third `GenCADFile::get_padstack_radius` - seems to be fixable in exactly same way like changes already present in the PR, since only text `contents` field used on `pad_name_ast`.

galkinvv:
I've tested the suggested changes with several .cad files that I use - all opens fine.
Also breifly reviewed the code - I don't see any problems.

So, this PR looks good for me!

About 3 methods mentioned:
- `GenCADFile::get_padstack_by_name` and `GenCADFile::get_pad_by_name` utilize the `GenCADFile::has_text_content` method. This method was introduced only as an extended string comparison that checks for string equality with graceful checking the first `mpc_ast_t*` argument for nullptr. No any "unobvious idea" was put in that method, just string comparison wrapper. So, if you change `pad_name` getting to use `get_nonquoted_or_quoted_string_child` - you can just swap `has_text_content` to "nullptr check + string comparison".
- The third `GenCADFile::get_padstack_radius` - seems to be fixable in exactly same way like changes already present in the PR, since only text `contents` field used on `pad_name_ast`.

swiftgeek:
Found and fixed another issue, this time with files created by CAMCAD:
https://github.com/swiftgeek/OpenBoardView/commit/f90784ba3f967f4477b53997e0385b7307a8889a

swiftgeek:
@galkinvv thanks, I think I know where I got confused now

`pad_name_ast` != `pad_ast`, and changing `pad_name_ast` does not affect `pad_ast` at all, so return remains valid and still of `mpc_ast_t` type. Initially I was under impression that this all would change and would not fit rest of the code. Thanks for pointing me in the right direction.

swiftgeek:
And made something horrible to deal with quotes, https://github.com/swiftgeek/OpenBoardView/commit/47a484f08421619fe9d7bd7f6a0ab2affd3fc416  but I can't tell that I know what I'm doing there, so won't push that to `gencad-rebased-upstream` unless somebody looks at it and it's fine.

swiftgeek:
I also tried fixing THT placement a bit, but this should be parsed for each pad individually - eg. many connectors have mix of SMT+THT pads, like some M.2 slots

https://github.com/swiftgeek/OpenBoardView/commit/9e6da140f67ce98f8c95ac069d4b456b1066f7f1

So as it is, that particular commit is not suitable for merging, and I don't think I will figure out how to control it for individual pads.

swiftgeek:
Got it working for individual pins after all with https://github.com/swiftgeek/OpenBoardView/commit/1623739e7203aa3a7c2f20a61b1ec31b4780c046

<details>
  <summary>Screenshots with STM32F0 disco board - M1076D that include through hole headers</summary>
  
 Expected board outline
board-outline-expected

Current OBV render:
board-outline-current
  
</details>

The issue I'm having now is that with component outline being on one side only, pins on other side cannot be selected with mouse, and I don't necessarily want to clutter viewer with component outlines on both sides as it has ability to make experience worse (like with files that include shield already)

inflex:
@swiftgeek OT - but do you have a link to the GenCAD file for the STM32F0 M1076D ?

swiftgeek:
ST has layout files publicly available and with some effort every layout file can be converted to GenCAD. Attaching M1076D GenCAD exported by `pcbnew` (KiCAD) and with artwork (traces) [stripped by my script](https://github.com/swiftgeek/pcbtools/blob/main/gencad/gencad_strip_art.sh)

[MB1076.zip](https://github.com/OpenBoardView/OpenBoardView/files/8964266/MB1076.zip)

As KiCAD adds new importers, things should be easier with time and so far I mostly tested it with P-CAD, Altium and [Cadence Allegro files after `extracta`](https://gitlab.com/kicad/code/kicad/-/issues/11826)

piernov:
@swiftgeek @galkinvv I'm sorry I haven't really followed what was happening with the GenCAD development. Is this ready to be merged?

swiftgeek:
I think it's ready

piernov:
Thanks again for the great work. I'm gonna release an Alpha1 if there are no other immediate changes to merge.

swiftgeek:
I think it would be a good idea to merge #155  as well, since GenCAD is providing rich descriptions for parts

PR #249, opened 2022-06-15
Improved interactive update of search-preview-highlighted parts.

galkinvv:
On clicking checkboxes/radiobuttons:
- added immediate update of search preview by currently focused column
- removed resetting of focused search column to first.
(search column is kept while changing searching parameters by storing its index in m_active_search_column)

On changing search column:
- added immediate update of highlighted parts (search preview by currently focused column).
- so tabbing from first column with text to the second empty column immediately clears preview.

On clearing text inside search column:
- added clearing of search preview. Before this patch the preview showed the results just before the clearing.

On opening popup again after performing search with Enter:
- added immediate previewing of text kept from previous search
- the focused column is kept, not reset to the first

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -1725,6 +1725,7 @@ void BoardView::SearchComponent(void) {
 		//		char cs[128];
 		const char *first_button[] = {m_search[0], m_search[1], m_search[2]};
 
+		bool search_params_changed = m_showSearch; //treat initiail popup reopening similar to later search option changes
 		if (m_showSearch) {
 			m_showSearch       = false;
 			m_tooltips_enabled = false;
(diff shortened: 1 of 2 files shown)


galkinvv:
Overall this PR tries to make interaction with search window more intuitive - the highlighted parts and nets always correspond to the currently active text field. Even if current text field is empty.

Also I just found that TAB/Shift-TAB navigation between text fields was broken in mainline before this patch, this is fixed as a bonus.

While improving intuitivity there is one situation that _may_ be treated as a behaviour worsening: with old logic switching to second text column left previous resulst highlited, so you can see what was found just before entering second column text.
However from my experience with this dialog the overall consistency seems more important that this "small feature".

The main reason I'm trying to make this dialog more consistent is plans to add one more search setting https://github.com/OpenBoardView/OpenBoardView/pull/155

I used that branch with extra search setting myself for several years now and start to fell that adding more features to search dialog requires making its behaviour more clear and intuitive. Otherwise the added feature would increase its complexity over the usable-by-new-users level

About the behaviour changes themselves.
They are summarized in commit message as text, however to make them more simple to undestand I prepared two screen capture videos - before and after the change

https://user-images.githubusercontent.com/1609972/173941952-ac7e5808-e410-40d1-bb57-721b000e1174.mp4

https://user-images.githubusercontent.com/1609972/173941948-dcd73ab7-d12e-4fba-b92c-20529ca8ec99.mp4

PR #255, opened 2022-07-18
Active-low label support from Kicad

Guimli:
Is it possible to support Active-low labels in Kicad format in the PIN/Net list:
[https://klc.kicad.org/symbol/s4/s4.7/](url)
But think it will be complicated to code...


piernov:
For feature requests, an issue should be opened at https://github.com/OpenBoardView/OpenBoardView/issues instead of a pull request.

PR #256, opened 2022-08-13
CMake: Search for stb through pkg-config, with find_path() fallback

davidebeatrici:


src/CMakeLists.txt
```diff
@@ -170,11 +170,21 @@ else() # Other compiler/version are unsupported, use external lib
...
 
 ## stb ##
-execute_process(
-	COMMAND git submodule update --init src/stb
-	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-)
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/stb)
+if(NOT WIN32 OR MINGW)
+	pkg_check_modules(STB REQUIRED stb)
+endif()
```
davidebeatrici, line 175: Maybe we should make `pkg_check_modules()` a no-op instead of performing a check every time we use it.

Alternatively: https://github.com/mumble-voip/mumble/blob/8fc359ea3fafbbb136abece6d422af409274b23e/cmake/pkg-utils.cmake

piernov:
Upstream stb doesn't provide a pkg-config configuration (and doesn't plan on it since it's not supposed to be used this way, but who cares) so pkg-config support is distribution-specific. To be more universal we should also try to look for the `stb_image.h` header as we do for `utf8.h`.

davidebeatrici:
You're right, I didn't realize it was a distro-specific addition.

I'll add a fallback (i.e. search for `stb_image.h` manually).

piernov:
`PATH_SUFFIXES stb` should be added to the `find_path()` command so that it is able to find the header in a `stb` subdirectory, such as `/usr/include/stb/stb_image.h`.

davidebeatrici:
Yep, I would create a macro/function.

piernov:
Looks good to me, thanks for the contribution.

davidebeatrici:
Thanks for merging!

Issue #258, opened 2022-08-21
GenCAD NB8513_MB_V3.cad $BOARD and $DEVICES parsing error


piernov:
Attached file does not open with master branch.
`$BOARD` does not get parsed properly due to the `TEXT` entry not in the grammar, so just a matter of adding it I guess.
`$DEVICES` does not get parsed properly due to `DEVICE LRC  LBAT54CWT1` having multiple unquoted words I think. No idea what this is about.

@swiftgeek or @galkinvv if one of you want to take a look at it.

[NB8513_MB_V3.zip](https://github.com/OpenBoardView/OpenBoardView/files/9389578/NB8513_MB_V3.zip)

swiftgeek:
Looks like newer CAMCAD (likely one included with PADS) started doing something unexpected, or input file got somewhat corrupted on import/export to product this strange result

piernov:
Do not hesitate to post a new comment rather than editing your previous one if you have new information, we do not get notified of edits.

If you think this really is a malformed file we could simply edit it to fix it, but if there's a possibility there are other files out there with the same kind of issues it's best for the parser to handle these cases.
Also, in general it would be better for the parser to be more permissive but I assume it's not easy with a grammar-based parsing.

As for the actual parser code, I cannot provide any insight unfortunately. Maybe @galkinvv can share some ideas if he can spare some time to look at it.

Otherwise, if you think your changes are "good enough" to handle the various files we've seen until now we can merge that. If it breaks we can fix it in a future beta release (although it seems like nobody really test alpha or beta releases).

swiftgeek:
I would rather want IRC channel on libera or oftc working like before, if possible ^^

CAMCAD is probably the most used used exporter (though maybe an older version of it), so I will need to test later what exactly is to blame. GenCAD itself is pretty widely used though, so that divergence from spec is quite disturbing.

GenCAD support is pretty important to me and the last thing I would want is it being broken/regressing, especially when I'm not sure what exactly am I doing to the code.

galkinvv:
`GenCADFile::get_device_by_name` was added by @martonmiklos during creating of first WIP prototype. Maybe there were some plans for using it, but now gencad renders well without it, so I think deleting it is safe by now.

About regressions - the best idea I have now - have a set of dome different cad files to be able at least test manually.
Some such files has "shady" source, so I'm not sure that attaching them to the projects github is a good idea.

I've created a google shared folder with them (and a subfolder with files having rendering problems).
[...]

PR #259, opened 2022-08-21
GenCAD: strip quotes from DEVICE field

swiftgeek:
~~Unreviewed~~ leftover commit from #247

KiCAD GenCAD files use a lot more quotes, and `DEVICE` field quote makes its way into OBV's description of component

GenCAD file for reference: [S3-OLinuXino_Rev_A.zip](https://github.com/OpenBoardView/OpenBoardView/files/8972216/S3-OLinuXino_Rev_A.zip)

@martonmiklos  @galkinvv

---

Related issue that could make situation even worse #258 , but actually it's fixed with two more commits, one changing parser a bit to deal with this new CAMCAD quirk, and other that removes dead code so we don't need to adjust it due to parser changes - https://github.com/swiftgeek/OpenBoardView/tree/gencad-device-quotes-camcad

This could be part of this pull or separate pull depending on reviewer's input


src/openboardview/FileFormats/GenCADFile.cpp
```diff
@@ -599,14 +599,22 @@ bool GenCADFile::has_text_content(mpc_ast_t *content_holder, const char *text) {
...
 	mpc_ast_t *ret_ast = mpc_ast_get_child(parent, key);
 	if (ret_ast) {
 		free(key);
 		return ret_ast->contents;
 	}
 
+	sprintf(key, "%s|wrapper_to_end|string|>", name);
+	ret_ast = mpc_ast_get_child(parent, key);
+	if (ret_ast) {
+		free(key);
```
galkinvv, line 613: this free in uncondiional, but return below is conditional

So when return is not done - use after free problem  in `sprintf(key, "%s|string|>", name);`

galkinvv: And yes, similar problem with doule-free in the next if in already existing (partially mine) code...

swiftgeek: So removing those two `free(key);` should fix it?
It doesn't seem to explode without them.

swiftgeek: [I think there is one more in double-free in `GenCADFile::get_nonquoted_or_quoted_string_child`](https://github.com/swiftgeek/OpenBoardView/commit/5090503f0b27c9d2ceaab92ae06d795763e75b37)

galkinvv: No, such change would lead to leak.

Without refactoring or paradigm change the concept should be "free just before every return" - something like this
```
	sprintf(key, "%s|wrapper_to_end|string|>", name);
	ret_ast = mpc_ast_get_child(parent, key);
	if (ret_ast) {
		auto value_ast = mpc_ast_get_child(ret_ast, "regex");
		if (value_ast) {
			free(key);
			return value_ast->contents;
		}
	}

	sprintf(key, "%s|string|>", name);
	ret_ast = mpc_ast_get_child(parent, key);
	if (ret_ast) {
		auto value_ast = mpc_ast_get_child(ret_ast, "regex");
		if (value_ast) {
			free(key);
			return value_ast->contents;
		}
	}
	free(key);
	return nullptr;
```

With refactoring/paradign change - can be many alternatives - changing execution path to the single return point, using C++ memory management via string or via std::unique_ptr.

The choice between "minimal change" VS "with refactoring" is really hard... unless your project owner.

galkinvv: >I think there is one more in double-free in GenCADFile::get_nonquoted_or_quoted_string_child
You are right. Same double-free there too.

swiftgeek: [I can still somewhat grasp that minimal change](https://github.com/swiftgeek/OpenBoardView/commit/fd19907c01a10c23ffb608ff0f979c93b4ca97dd) with one more place in `GenCADFile::get_nonquoted_or_quoted_string_child` (this time better I hope)

galkinvv: looks great! (this PR can be updated with the change).

swiftgeek:
I guess then it's ready for merge?
@piernov

piernov:
In the future it'd be better to try to use C++'s STL for string manipulation and containers since using `malloc`, `free` and `sprintf` (`snprintf` would already be a little bit better) can easily lead to catastrophic mistakes. Although I know the existing code is a bit of a mess.

I think this should be good though so let's go ahead.
Thanks again for the contribution.

Issue #248, opened 2022-06-14
GenCAD board outline issue


swiftgeek:
(originally posted in https://github.com/OpenBoardView/OpenBoardView/pull/219#issuecomment-1147291206)
I have prepared a file with a board outline that still has issues loading properly
[board-outline.zip](https://github.com/OpenBoardView/OpenBoardView/files/8843636/board-outline.zip)

<details>
  <summary>Screenshots with expected behaviour/current state</summary>
  
 Expected board outline
![board-outline-expected](https://user-images.githubusercontent.com/270845/172141682-bd4d3172-8986-4b8d-84b6-a34ae75e4d15.png)

Current OBV render:
![board-outline-current](https://user-images.githubusercontent.com/270845/172142072-510eb770-4edc-4007-9054-752894a8590e.png)
  
</details>

piernov's response in pull request adding GenCAD support: https://github.com/OpenBoardView/OpenBoardView/pull/219#issuecomment-1147744835

> There was something similar happening in BVR3 format due to having a segmented outline mode which OBV doesn't know how to handle, no idea if it's related but you can look at the discussion in #224 and the code in "BVR3File.cpp".

inflex:
I suspect the problem is that the GenCAD outline usually is segment based, not a sequence of poly points. 

You've got two options that I can think of immediately;

1) Add segment-based outline + fill drawing versions of the existing outline/fill functions
2) Convert segment based outlines to polygon sequence ( note that internal holes will need to be handled either by "same point termination" or clockwise-anticlockwise path directions.

This is what it looks like in FlexBV ( which has the segment based rendering facility
![s](https://user-images.githubusercontent.com/5660591/173719711-ec389570-c27a-42de-924a-f3cf25550c5a.jpg)
 )

martonmiklos:
So here is an example gencad forming a circle contour with 100 radius around the 100,100 point:
https://gist.github.com/martonmiklos/27564ca2fc0ac7108c7fccc117e1413a

This renders this way:
![kép](https://user-images.githubusercontent.com/1609182/174071357-3e69ce5c-d22f-4b92-af72-c4084ea6286f.png)

PR #261, opened 2022-08-23
Segmented outline drawing mode for GenCAD

piernov:
Should mostly fix outline problems of GenCAD.
Hopefully I didn't break too many things
Breaks at least board fill but really who cares about that.
Also arcs beginning/end do not match exactly with the previous/next line, could be due to float/int conversions.

@swiftgeek @galkinvv

### DIFF
--- a/src/openboardview/FileFormats/GenCADFile.cpp
+++ b/src/openboardview/FileFormats/GenCADFile.cpp
@@ -7,6 +7,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <limits>
+#include <optional>
 
 #ifndef M_PI
 #define M_PI 3.14159265358979323846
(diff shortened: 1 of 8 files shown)


swiftgeek:
Besides broken board fill, I don't see downsides, tested on gencad files ranging from mobile devices with detailed complex outlines to strange server mobos.

src/openboardview/FileFormats/GenCADFile.cpp
```diff
@@ -410,88 +434,43 @@ bool GenCADFile::parse_board_outline(mpc_ast_t *board_ast) {
...
-			p.x = pc.x + r * cos(i);
-			p.y = pc.y + r * sin(i);
-			format.push_back(p);
-			num_format++;
-		}
-		format.push_back(p1);
-		num_format++;
+		pold = p;
 	}
+	arc_segments.push_back({p, p2});
```
galkinvv, line 471: This logic may lead to "artifact" segment from (0;0) if some board would contain arc with p1==p2, since in this case startAngle would be exactly equal to endAngle and for cycle would not run at all.

One of the "simple fix" variants may be initializing p with p1.

galkinvv:
I've reviewed at gencad-related chnages - it seems all-fine with 1 corner-case issue written above.

From ther "user point of view" I cant say anything, since never payed attention to outlines.

swiftgeek:
Well it's hard to ignore board outline when it looks like in #248 (or worse)
Luckily with GenCAD we can just trim things that cause problems like `$BOARD` in this case.

PR #262, opened 2022-08-24
GenCAD: Fix goldfinger parsing

swiftgeek:
An attempt to fix edge connector parsing (like DIMM/PCIe Card goldfinger), tested with attached test files

~~Also rename `GenCADFile::is_padstack_smd` to `GenCADFile::is_padstack_drilled` to avoid confusion~~ done

~~Most likely, non-copper/drill elements (like silkscreen only) being not omitted (last leftover issue from #247) is related to this change, but somehow padstack would need to be marked as empty (`BRDPinSide::None`) and then component omitted if it doesn't have any copper/drill elements (maybe `is_shape_smd()` would need to change to return part_type directly (so rename to `get_shape_type()`), and return `BRDPartType::None` if it doesn't have any useful elements). `get_shape_type()` could iterate similarly to `get_shape_type()` now, to determine there are no drill/outer copper elements in given shape. Would also make easier any future expansion of `BRDPartType` by via/testpoint or other types~~ implemented

---

[PantherPlus from OCP (Open Compute Project](https://github.com/OpenBoardView/OpenBoardView/files/9431811/PantherPlus.zip)

[Glasgow Debug Tool](https://github.com/OpenBoardView/OpenBoardView/files/9902483/glasgow.zip) - Due to KiCAD limitation, 8 pads from J5 are composed out of several pads each

### DIFF
--- a/src/openboardview/FileFormats/BRDFileBase.h
+++ b/src/openboardview/FileFormats/BRDFileBase.h
@@ -36,7 +36,7 @@ struct BRDPoint {
 };
 
 enum class BRDPartMountingSide { Both, Bottom, Top };
-enum class BRDPartType { SMD, ThroughHole };
+enum class BRDPartType { None, SMD, ThroughHole };
 
 struct BRDPart {
 	const char *name = nullptr;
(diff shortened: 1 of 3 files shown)


piernov:
Can you provide an example to what issue this addresses? (file/screenshot/whatever)

swiftgeek:
[PantherPlus from OCP (Open Compute Project](https://github.com/OpenBoardView/OpenBoardView/files/9431811/PantherPlus.zip)

swiftgeek:
Ready for review

piernov:
Superseded by #269 using some commits from this PR. Thanks for the contribution.

PR #263, opened 2022-09-05
PDFBridgeSumatra: Detect use of fork on runtime

swiftgeek:
Not even using windows, and I have not seen build of the forked sumatra yet.

But tested with official 3..4.6 and PDF search works!

### DIFF
--- a/src/openboardview/PDFBridge/PDFBridgeSumatra.cpp
+++ b/src/openboardview/PDFBridge/PDFBridgeSumatra.cpp
@@ -139,6 +139,38 @@ bool PDFBridgeSumatra::InitializeDDE() {
 	return true;
 }
 
+bool PDFBridgeSumatra::TestFork() {
+	if (this->ForkTested)
+		return true; // Already determined
+
+	if (this->idInst == 0L) {
+		SDL_LogError(SDL_LOG_CATEGORY_ERROR, "%s", "PDFBridgeSumatra DDE not initialized");
+		return false;
+	}
+
+	if (this->hConv == nullptr) {
+		if (!this->ConnectDDEClient(this->szService, this->szTopic)) {
+			return false;
+		}
+	}
+
+	std::wstring ddeCmd = L"[Search(\"" + this->pdfPathString + L"\",\" 0p3nb04rdv13w \",0)]";
+	if ( this->ExecuteDDECommand(ddeCmd)) {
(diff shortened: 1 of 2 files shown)


src/openboardview/PDFBridge/PDFBridgeSumatra.cpp
```diff
@@ -139,6 +139,38 @@ bool PDFBridgeSumatra::InitializeDDE() {
 	return true;
 }
 
+bool PDFBridgeSumatra::TestFork() {
+	if (this->ForkTested)
+		return true; // Already determined
```
MattSturgeon, line 144:
```suggestion
		return this->IsForked; // Already determined
```

Why can't this return `this->IsForked`? You don't currently use the return value when calling this method, but to me it makes sense to use this method to check if Sumatra is forked :thinking: 

e.g. `if (this->TestFork()) { /* do stuff */ }`

Should probably document what the return value is for if it's for something else (like indicating an error state)

src/openboardview/PDFBridge/PDFBridgeSumatra.cpp
```diff
@@ -309,7 +341,12 @@ void PDFBridgeSumatra::DocumentSearch(const std::string &str, bool wholeWordsOnl
 	}
 
 	auto wstr = utf8_to_utf16(strCopy);
-	std::wstring ddeCmd = L"[Search(\"" + this->pdfPathString + L"\",\"" + wstr + L"\"," + (caseSensitive ? L"1" : L"0") + L")]";
+	std::wstring ddeCmd;
+	this->TestFork();
+	if (this->IsForked)
+		ddeCmd = L"[Search(\"" + this->pdfPathString + L"\",\"" + wstr + L"\"," + (caseSensitive ? L"1" : L"0") + L")]";
+	else
+		ddeCmd = L"[Search(\"" + this->pdfPathString + L"\",\"" + wstr + L"\")]";
```
MattSturgeon, line 349: This could be simplified to `if (this->TestFork());` if the above suggestion is taken. Not sure how I feel about having the assignment nested in an `if..else` vs having an additional ternary in the assignment... I think split over two or three lines it's still fairly readable?

```suggestion
	std::wstring ddeCmd = L"[Search(\"" + this->pdfPathString + L"\",\"" + wstr + L"\""
		+ (this->TestFork() ? L"," + (caseSensitive ? L"1" : L"0") : L"") + L")]";
```

MattSturgeon (commented):
I'm a bit rusty on C++, so I may say something silly. But I've had a quick once over and left some comments.

PR #276, opened 2023-03-05
GenCADFile: don't duplicate shape name if it is included in device name

galkinvv:
Some .cad files have identical data in shape and device name. Duplicating it makes display overloaded.
This change skips shape name display if it is contained in device name.

### DIFF
--- a/src/openboardview/FileFormats/GenCADFile.cpp
+++ b/src/openboardview/FileFormats/GenCADFile.cpp
@@ -203,10 +203,13 @@ bool GenCADFile::parse_components() {
 			if (shape_ref_ast) {
 				char *shape_name_str = get_nonquoted_or_quoted_string_child(shape_ref_ast, "shape_name");
 				if (shape_name_str && *shape_name_str) {
-					if (!brd_part.mfgcode.empty()) {
-						brd_part.mfgcode += " SHAPE ";
+					if (brd_part.mfgcode.find(shape_name_str) == std::string::npos) {
+						// cat shape name to displayed information if it is not already included
+						if (!brd_part.mfgcode.empty()) {
+							brd_part.mfgcode += " SHAPE ";
+						}
+						brd_part.mfgcode += shape_name_str;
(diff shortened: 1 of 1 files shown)


galkinvv:
I have a .cad Asus(?) GPU board which contains a lot of elements where device name is identical to a shape name, maybe some conversion result. For this board the duplication in disaplyed names like `MULTI_DFET_010_SMD_050X060 SHAPE MULTI_DFET_010_SMD_050X060` seems overloaded to me.

However, the suggested change may be opinionated, since strictly speaking "duplicated shape information is also an information".

So I'm fine if this will be closed/rejected)

galkinvv:
Rebased PR on latest beta & checked that it builds&works.
Without this patch _some_ gencad files produces too much duplication in part information.

Here is the example file that motivated me to adjust this:
https://drive.google.com/file/d/10Ky7a25J8FEZC0-lHYsvcc3qUHAlPMxi/view?usp=drive_link

Without this patch the Q13 is named as 
`q_fet_n_enh_MULTI_DFET_010_SMD_050X060 SHAPE MULTI_DFET_010_SMD_050X060`

image


The patch shorters this to `q_fet_n_enh_MULTI_DFET_010_SMD_050X060` because the info after SHAPE is just a duplication for this part.

piernov:
With this patch the "SHAPE" delimiter is lost for parts that have this information duplicated, so it makes it inconsistent with parts that don't. The "SHAPE" delimiter in itself was arbitrary anyway, but I think it's better to keep it consistent across all parts of a board. And without the delimiter you could wonder if the shape information is actually missing as it may not be clear it is already included in the part name.
Another way would be to remove the shape information from the part name and keep it only after the shape delimiter, but then weird things could happen if somehow the part name contains the shape as a substring as part of the actual part number (e.g., an "SMD" shape and a part number "SMD123456X").

As a side note, ideally the shape should not be appended to the part name string but rather put into another field of the part structure.

PR #277, opened 2023-03-06
FZFile: parse pad names, enable pad sorting

galkinvv:
Some .fz files contains pad names,
this change enables their parsing and display.

With this information available the pin sorting is enabled.

If the pin numbers are identical those names are also used for sorting: a special ordering function is implemented to order pins of large BGAs as A0 A1..A9 A10 A11..A99 A100 A101..B0..Z0..AA0..AZ0..BA0..

Unfortunately some .fz files has ICs with very long pin names like VINSEN_AUX_1; such long names makes display cluttered.

### DIFF
--- a/src/openboardview/FileFormats/BRDFileBase.cpp
+++ b/src/openboardview/FileFormats/BRDFileBase.cpp
@@ -2,6 +2,7 @@
 
 #include "utf8/utf8.h"
 #include <cstdint>
+#include <cstring>
 
 // from stb.h
 void stringfile(char *buffer, std::vector<char*> &lines) {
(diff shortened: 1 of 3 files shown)


galkinvv:
After year of testing this locally in aday-to-day use I have to convert this back to draft - there is a regression for some files and I didn't find a clear solution yet.

This comment is meant as a notes useful for further enhcancements of FZ files (for me in the future or anybody reading this).

It turns out that there at least 2 variants for information attached to pins in commonly found fz boardviews.
Uploaded to _FZ-samples_ folder for reference at https://drive.google.com/drive/folders/1XD3OSYpvj5znjMmG58cHdOut3f2ltg--?usp=drive_link

Variant for AsRock motherboard has bga pad position `AJ43` in column PIN_NUMBER
```
A!NET_NAME!REFDES!PIN_NUMBER!PIN_NAME!PIN_X!PIN_Y!TEST_POINT!RADIUS!
...
S!SATA_GP1!SU1!AJ43!GPP_E1/SATAXPCIE1/SATAGP1!3315.86!1830.70!!6!
```

And variant for Strix GPU always has "0" in column PIN_NUMBER and bga pad position `AJ31` in column PIN_NAME
```
A!NET_NAME!REFDES!PIN_NUMBER!PIN_NAME!PIN_X!PIN_Y!TEST_POINT!RADIUS!
...
S!SNN_FBA_WCKB45*!G1!0!AJ31!3140.51!1436.167!!7.48106!
```

Those can be handled by skipping 0 in PIN_NUMBER columns but further handling of the extra info from pin role can be complicated. I thought about ways to handle it while keeping the code supportable and came to a conclusion that it may be handled via a some terminology adjustment.

By now pins a defined to have `number/snum` - ment to be the index for a numbered nails and a `name` - meant to be a BGA pad name like AZ82. The new info available in FZ file is not like number/identifier it is "pad role of a package". While being related - it IS different from net it is connected to. Typically it is the pad name from "isolated/generic" part datasheet, not net name from board corresponding to a specific usage inside a board.

So I came to an idea of unifiying current number and name in a single variable like `position_id`, and introducing new field of a pad strust named like "role_in_package" for handling such info. Unifiying name&number fields is a precondition to this since having another one string filed would make visualizing it even more complicated

galkinvv:
Closing, this is effectively split into https://github.com/OpenBoardView/OpenBoardView/pull/307 and https://github.com/OpenBoardView/OpenBoardView/pull/361

PR #280, opened 2023-03-28
GenCADFile: add support for GenCAD files generated by "XY html to CAD1.2"

galkinvv:
This GenCAD variant has following differences from previously supported GenCAD files:
 - lack of $BOARD, $TRACKS, $ROUTES sections
 - signal names with spaces may be non-quoted
 - extra spaces at the end of NODE of SIGNAL


galkinvv:
I've encountered a GenCAD variant containing notice that it was created with "XY html to CAD V1.2" utility.

It turns out that such file had some incompatibilities with OpenBoardView, this PR tries to resolve them.

ALso I'm attaching the minified to 1 element board as a test file for this format
[xyhtml-to-cad-result-minimal-sample.cad.zip](https://github.com/OpenBoardView/OpenBoardView/files/11094574/xyhtml-to-cad-result-minimal-sample.cad.zip)

PR #281, opened 2023-03-31
Allow system mpc

Apteryks:
Hi,

This enables building OpenBoardView using a system-provided MPC library. 

Thanks!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -206,13 +206,14 @@ endif()
 add_definitions(-DSTBI_FAILURE_USERMSG)
 
 ## mpc - Micro Parser Combinators ##
-execute_process(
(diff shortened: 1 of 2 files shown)


piernov:
Due to the name conflict with `libmpc`, which is vastly more common and usually installed in most build environments, I'm not sure it's a good idea to rely on this.
If the renaming of `mpc` (Micro Parser Combinators) to `pcq` actually became the default we could use that instead, but it seems like it's abandoned.

Apteryks:
The name conflict with GNU mpc is annoying, but given that GNU mpc doesn't ship any pkg-config file, one can be sure that 'mpc' found via pkg-config would be the orangeduck-mpc (that's how I named the package in GNU Guix to avoid the name clash).

PR #307, opened 2024-07-28
FZFile: parse PIN_NAME if PIN_NUMBER string is useless

galkinvv:
This is a basic improvement for handling .fz files having useless values in PIN_NUMBER.

There are dozens of .fz board files, maybe all converted by same tool having **all** PIN_NUMBER strings as a single-digit "0" but having actual pin numbers and names in PIN_NAME column.

This patch is a minimal fix that enables displaying such data without a huge refactoring

(initially I tried a more complete attempt in https://github.com/OpenBoardView/OpenBoardView/pull/277 but failed and came to a conslusion that at least first several steps should be done by a smaller PRs)

### DIFF
--- a/src/openboardview/FileFormats/FZFile.cpp
+++ b/src/openboardview/FileFormats/FZFile.cpp
@@ -357,12 +357,29 @@ FZFile::FZFile(std::vector<char> &buf, uint32_t fzkey[44]) {
 				parts_id[part.name] = parts.size();
 			} break;
 			case 2: { // Pins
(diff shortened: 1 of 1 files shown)


galkinvv:
Example of such file is the Strix GPU board in https://drive.google.com/drive/folders/1PQKd-OrDm6DgtpAa3eTy7-pTA7yNTQZF?usp=drive_link

piernov:
Hopefully there are no situation where the digit 0 is used as an actual pin number.

PR #309, opened 2024-12-03
Initial XZZ .pcb file support

slimeinacloak:
Basic file format support for XinZhiZao's .pcb format
Also I have added a comment data type to the pin struct which is used to display diode readings when hovering

Notes on the format:
- All their files are flat so flipping the board does not work and there isnt a clean way to unmirror the files
- Part details are included in some files but I haven't fully decoded those sections so that information is not displayed at the moment

### DIFF
--- a/src/openboardview/FileFormats/BRDFileBase.h
+++ b/src/openboardview/FileFormats/BRDFileBase.h
@@ -64,6 +64,7 @@ struct BRDPin {
 	double radius    = 0.5f;
 	const char *snum = nullptr;
 	const char *name = nullptr;
+	const char *comment = nullptr;
 
 	bool operator<(const BRDPin &p) const // For sorting the vector
 	{
--- a/src/openboardview/FileFormats/XZZPCBFile.cpp
+++ b/src/openboardview/FileFormats/XZZPCBFile.cpp
@@ -1,0 +1,890 @@
+#include "XZZPCBFile.h"
+#include "utils.h"
+#include <cctype>
+#include <clocale>
+#include <cstdint>
+#include <cstring>
+#include <vector>
+#include <cstdint>
+#include <numeric>
+#include <algorithm>
+#include <list>
+#include <iostream>
+#include <string>
+#include <fstream>
+#include <vector>
+#include <iomanip>
+#include <unordered_map>
+#include "annotations.h"
+#include <vector>
+#include <cstring> // For memcpy
+#include <cctype>  // For isprint
+#include <cstdio>  // For fprintf
+#include <cstdint> // For uint64_t
+#include <iostream>
+#include <iomanip>
+
+#define LB32_MASK   0x00000001
+#define LB64_MASK   0x0000000000000001
+#define L64_MASK    0x00000000ffffffff
+#define H64_MASK    0xffffffff00000000
+
+/*
+ * Credit to @huertas for DES functions
+ * Also credit to @inflex and @MuertoGB for help with cracking the encryption + decoding the format
+ */
+
+/* Initial Permutation Table */
+static char IP[] = {
+    58, 50, 42, 34, 26, 18, 10,  2, 
+    60, 52, 44, 36, 28, 20, 12,  4, 
+    62, 54, 46, 38, 30, 22, 14,  6, 
+    64, 56, 48, 40, 32, 24, 16,  8, 
+    57, 49, 41, 33, 25, 17,  9,  1, 
+    59, 51, 43, 35, 27, 19, 11,  3, 
+    61, 53, 45, 37, 29, 21, 13,  5, 
+    63, 55, 47, 39, 31, 23, 15,  7
+};
+
+/* Inverse Initial Permutation Table */
+static char PI[] = {
+    40,  8, 48, 16, 56, 24, 64, 32, 
+    39,  7, 47, 15, 55, 23, 63, 31, 
+    38,  6, 46, 14, 54, 22, 62, 30, 
+    37,  5, 45, 13, 53, 21, 61, 29, 
+    36,  4, 44, 12, 52, 20, 60, 28, 
+    35,  3, 43, 11, 51, 19, 59, 27, 
+    34,  2, 42, 10, 50, 18, 58, 26, 
+    33,  1, 41,  9, 49, 17, 57, 25
+};
+
+/*Expansion table */
+static char E[] = {
+    32,  1,  2,  3,  4,  5,  
+     4,  5,  6,  7,  8,  9,  
+     8,  9, 10, 11, 12, 13, 
+    12, 13, 14, 15, 16, 17, 
+    16, 17, 18, 19, 20, 21, 
+    20, 21, 22, 23, 24, 25, 
+    24, 25, 26, 27, 28, 29, 
+    28, 29, 30, 31, 32,  1
+};
+
+/* Post S-Box permutation */
+static char P[] = {
+    16,  7, 20, 21, 
+    29, 12, 28, 17, 
+     1, 15, 23, 26, 
+     5, 18, 31, 10, 
+     2,  8, 24, 14, 
+    32, 27,  3,  9, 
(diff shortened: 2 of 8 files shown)


src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+		exit(1);
+	}
+	
+	current_pointer += 31;
+	uint32_t part_name_size = *reinterpret_cast<uint32_t*>(&buf[current_pointer]);
+	current_pointer += 4;
+	std::string part_name(reinterpret_cast<char*>(&buf[current_pointer]), part_name_size);
+	current_pointer += part_name_size;
+
+	part.name = strdup(part_name.c_str());
```
piernov, line 699: This leaks memory, however we may not have a better way to deal with it right now. Would probably need to convert `BRDPart::name` to an `std::string`.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+				current_pointer += 4;
+				pin.pos.y = *reinterpret_cast<uint32_t*>(&buf[current_pointer]) / 10000;
+				current_pointer += 4;
+				current_pointer += 8; // currently unknown
+
+				uint32_t pin_name_size = *reinterpret_cast<uint32_t*>(&buf[current_pointer]);
+				current_pointer += 4;
+				std::string pin_name(reinterpret_cast<char*>(&buf[current_pointer]), pin_name_size);
+				
+				pin.name = strdup(pin_name.c_str());
```
piernov, line 745: Same comment about leaking memory here.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+				uint32_t net_index = *reinterpret_cast<uint32_t*>(&buf[current_pointer]);
+				current_pointer = pin_block_end;
+
+				std::string diode_reading = "";
+
+				std::string pin_net = net_dict[net_index];
+				
+				if (pin_net == "NC") {
+					pin.net = strdup("UNCONNECTED");
```
piernov, line 758: Unnecessary leaky `strdup()` since this is a constant.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+				} else {
+					pin.net = strdup(pin_net.c_str());
+				}
+				
+				pin.part = parts.size() + 1;
+
+				if (diode_reading != "") {
```
piernov, line 765: `diode_reading` is defined as empty string just above so this is not doing anything.

src/openboardview/FileFormats/BRDFileBase.h
```diff
@@ -64,6 +64,7 @@ struct BRDPin {
 	double radius    = 0.5f;
 	const char *snum = nullptr;
 	const char *name = nullptr;
+	const char *comment = nullptr;
```
piernov, line 67: Probably better to declare as an `std::string` already to avoid some memory management issues.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+					pin.comment = strdup(diode_reading.c_str());
+				} else if (diode_readings_type == 1) {
+					if (diode_dict.find(part.name) != diode_dict.end() && diode_dict[part.name].find(pin.name) != diode_dict[part.name].end()) {
+						pin.comment = strdup(diode_dict[part.name][pin.name].c_str());
```
piernov, line 769: If `BRDPin::comment` is declared as `std::string` the leaky `strdup()` becomes unnecessary, and the `c_str()` too.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+					}
+				} else if (diode_readings_type == 2) {
+					if (diode_dict.find(pin.net) != diode_dict.end()) {
+						pin.comment = strdup(diode_dict[pin.net]["0"].c_str());
```
piernov, line 773: Is `"0"` always present? Are there other possible keys?

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+					}
+				} 
+
+				pins.push_back(pin);
+				pin = blank_pin;
```
piernov, line 778: Can assign `{}` instead of having to declare `blank_pin`.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+				break;
+			}
+			default:
+				if (sub_type_identifier != 0x00) {
+					printf("Unknown sub block type: 0x%02X at %d in %s\n", sub_type_identifier, current_pointer, part_name.c_str());
```
piernov, line 784: Please use `SDL_Log…` to log messages to the console (without aborting file loading, otherwise ENSURE_OR_FAIL should be used).

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+	part.mounting_side = BRDPartMountingSide::Top;
+	part.part_type = BRDPartType::SMD;
+	
+	pin.snum = strdup(name.c_str());
+	pin.side = BRDPinSide::Top;
+	pin.pos.x = static_cast<double>(x_origin) / 10000;
+	pin.pos.y = static_cast<double>(y_origin) / 10000;
+	if (net_dict.find(net_index) != net_dict.end()) {
+		if (net_dict[net_index] == "UNCONNECTED" || net_dict[net_index] == "NC") {
+			pin.net = strdup(""); // As the part already gets the kPinTypeTestPad type if "UNCONNECTED" is used type will be changed to kPinTypeNotConnected
```
piernov, line 827: No need for `strdup()` since it is a constant.

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+            std::vector<uint8_t> test_pad_data((uint8_t*)(block_buf.data()), (uint8_t*)(block_buf.data() + block_buf.size()));
+            parse_test_pad_block(test_pad_data);
+            break;
+        }
+        default:
+            break;
+    }
+}
+
+char XZZPCBFile::read_utf8_char(char c) {
```
piernov, line 513: Dead code?

src/openboardview/FileFormats/XZZPCBFile.cpp
```diff
@@ -0,0 +1,890 @@
...
+    if (static_cast<unsigned char>(c) < 128) {
+        return c; // Print ASCII character as is
+    } else {
+        return '?'; // Replace non-ASCII character with '?'
+    }
+}
+
+std::string XZZPCBFile::read_cb2312_string(const std::string &str) {
```
piernov, line 521: Is it GB2312? Maybe it's possible to use `std::codecvt` to convert properly?

piernov (changes requested):
Thanks for the contribution.
I wrote some comments in relation with some parts of the code.

Additionally, please fix the indentation. OpenBoardView's code base uses a single tab for each indent level, as configured by the EditorConfig file. There is also a clang-format configuration to fix the styling automatically, but please double check its effect on the code.

For the FZ format, we had decided not to include the decryption key in the source code due to possible intellectual property infringement (see previous legal issues regarding keys in the multimedia and video game copy protection world), and once it is included in the Git history, it cannot be removed easily. To stay consistent, we would need to do the same here.

PR #315, opened 2025-02-24
Automated building via GitHub actions

galkinvv:
Introduce automated building via github actions instead of outdated & non-working travis.

Build target a variety of variants to ensure future- and backward- OS version compatitbility.
Thise builds can be useful for users trying to try prelimiary/PR features
(especially on windows where building is moe complex)

And for developers too see if they introduce any platform incompatibilities.


galkinvv:
To make this actually working github actions needs to be enabled for repository (if not yet).

Too see how it would be looking - see here https://github.com/galkinvv/OpenBoardView/actions?query=branch%3Agithub-action-builders 

Example artifact list:
```
openboardview-32bit.exe.15ead286ab3169ce2c14cf930a4c42a78f410be6 2.21 MB	
openboardview-ucrt64.exe.15ead286ab3169ce2c14cf930a4c42a78f410be6 2.62 MB	
openboardview_DEBUG-32bit.exe.15ead286ab3169ce2c14cf930a4c42a78f410be6 10 MB	
openboardview_DEBUG-ucrt64.exe.15ead286ab3169ce2c14cf930a4c42a78f410be6 8.34 MB	
openboardview_DEBUG_linux_elf.15ead286ab3169ce2c14cf930a4c42a78f410be6 7.01 MB	
openboardview_package_linux_deb.15ead286ab3169ce2c14cf930a4c42a78f410be6 1.36 MB	
openboardview_package_linux_rpm.15ead286ab3169ce2c14cf930a4c42a78f410be6 1.35 MB	
openboardview_package_macos13_intel_dmg.15ead286ab3169ce2c14cf930a4c42a78f410be6 1.75 MB	
openboardview_package_macos15_universal_dmg.15ead286ab3169ce2c14cf930a4c42a78f410be6 3.11 MB
```

Note: **the Mac .dmgs were not tested by me**, so while they are building - there is a chance that they are broken.
However thay were included to at least test buildability for Mac.

I've manually tested windows builds, .deb build, and even .rpm compatibility with CentOS8



I'm open to any comments/suggestions/reviews !

galkinvv:
One of the purposes of this is "simplify providing windows binaries for users interested in specific feature testing"

To use this on forks - you need enable Actions for forks (maybe disabled by default). After merging this to upstream - the builds should show up in upstream PRs too.

One of the purposes of getting this automation was "try to get windows build" of the XZZ branch.

The result is here: https://github.com/galkinvv/OpenBoardView/actions/runs/13488137928 

While 32-bit windows build failed, and MacOS build failed too (due to some incompatibilities in the new code) - the 64-bit windows binary is priducde and seems working fine (checked with X1799-820-02020-08.pcb)!
@slimeinacloak - you may be interested in this working windows binary of yours branch/list of platform compiler error incompatibilities.

Windows binary present in the artifacts, github account needed to download it. 
Note: the artifacts are not releases, theay are stored only for 3 months as far as I know from another project.

piernov:
Looks good overall, thanks, few remarks/questions below.

1. Could we separate the MinGW jobs and Docker image from the DEB/RPM build?
For DEB/RPM, it is indeed important to build on an "older" distribution since it is linked dynamically so that we avoid depending on a newer glibc. I was using Debian 10.13 for the DEB package and RHEL 8.3 for the RPM package.
For MinGW, the underlying distribution can be newer since the toolchain is independent and it is linked statically anyway. That way we avoid depending on a random outdated PPA, and the image for DEB/RPM doesn't get polluted by that.

2. As a side note for the previous point, I think we should try to provide an OBV build with a UCRT build of MinGW since that should provide better support for Unicode. MinGW 12 should default to UCRT and is available on Ubuntu 24.10.

3. Do we really need to start the container manually instead of relying on the container syntax of GitHub actions?

4. I think we need to fetch the entire repo (not just last commit) with:
```
        with:
          fetch-depth: 0
```
in the checkout action to get the correct version numbering in the build.

5. macOS build will not be signed so it won't run on Apple Silicon I'm pretty sure. I don't know how that's supposed to be handled with GitHub actions, and I don't have a signing certificate myself, I had someone else build it for me for the latest releases.

galkinvv:
Thank for comemnts

1. Adjusted this in added commits, now Dockerfile contains separate targets for "native linux build on older distribution" and for "mingw with its different requirements"
2. Practical experiments with cross compiling gave quite a low success rate:
    * prebuilt mingw sdl2 provided by sdl2 developers does not have lib build in UCRT64 mode. So I see no way for getting UCRT64 cross-builds for now without manually scripting sdl2 build.
    * To get things more complex I tried several versions of g++-mingw-w64-i686-posix and find that:
      * all 12.x/13.x/14/x crosscompiling I tried from debian/ubuntu/alpine fails with [linking error](https://github.com/galkinvv/OpenBoardView/actions/runs/13510552720/job/37749753486#step:7:915) multiple difinition of `.weak._utf8*.__ZSt23__is_constant_evaluatedv'. Maybe its real problem in utf-8 lib, but my intention is mostly "try building existing code as is, its pretty compatible really"
      * versions <10.3.0 has a bug in filesystem::rename that throws error if target file exists. This crashes openboardview on start due to conf renaming. [Gcc fixed this in 10.3.0](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=8dd1c1085587c9f8a21bb5e588dfe1e8cdbba79e)
      * I failed to find a lts-like distro providing mingw 11.x with expected host+target.
      * So, only 10.3.0 is a known-to-work version. Fortunately it was shipped with ubuntu 22.04.
    * However, the successful UCRT-build **IS** done via msys2's UCRT64 environment execeuted on windows - which provides its own sdl2 build.
 3. I didn't find a standard method to use just-build-in previous step container with caching (using cache speedups pipelines after the first makes it independent on external-to-github resources). This can be done via pushing containers to gcr.io, but this would not be asealy applicable to forks and require some manual cache mantaince.
 4. Thanks, now I undestood how this is counted and looks that version number is fixed in [latest build](https://github.com/galkinvv/OpenBoardView/actions/runs/13511982810)
 5. I added the "unsigned" in the name of Mac artifacts to make such behaviour more expectable by dowsnloaders. Action artifacts can't be directly downloaded by users without github account, so hope there would be some degree of problem understanding/knowledge how to circumvent this from people having github account.

piernov:
Thanks again for the work.

2. Thanks for the investigation, getting a MinGW build to work can indeed be a pain. I typically build locally with the latest MinGW available on ArchLinux (and MinGW SDL2 built from the AUR), so for the next release I'll probably have to find a way to fix the build issue. I missed the fact that there is a UCRT64 build on Windows so that's good.

PR #323, opened 2025-04-13
CMake: fix compatibility with CMake 4.x on mingw-w64-ucrt

galkinvv:
Support for CMake policies below 3.10 is deprecated since CMake 3.31 https://cmake.org/cmake/help/v4.0/command/cmake_minimum_required.html#policy-version

The oldest platform targeted by OpenBoardView is RHEL8-like distros, having CMake 3.11.

So switch to requitring CMake 3.11 - this automatically switches all older policies to NEW behaviour, so explicit policy seting is removed

Explicit declaration of needed version for subfolders is removed

Note: zlib submodule is not updated for CMake4 support,
so platforms having CMake4 are expected to provide system zlib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
-# We now use project() VERSION parameter
-if(POLICY CMP0048)
-	cmake_policy(SET CMP0048 NEW)
-endif()
-# Targets with semicolon are imported targets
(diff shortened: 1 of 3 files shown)


galkinvv:
I'm really sorry, while the 64-bit windows build succeeds now, the resulting binary doesn't work without zlib1.dll

I'm trying to fix the issue & hope to prepare another PR.

PR #325, opened 2025-04-28
Fix non-operational File->Open with macOS Sequoia.

inflex:
Due to the new security model, Sequoia now fails to launch the File-Open window because an inDomain value is null/nil.

Cause seems to be due to CMake using its own internal Info.plist generator which doesn't appear to work right for the new requirements.

Dedicated  Info.plist added in to asset folder. 

*** Consider this as a placeholder for people wanting to resolve the issue.  I will experiement to see if modifying the Info.plist parameters within CMake itself will resolve the issue

( try these options set within the CMakeList.txt instead of this PR fix )
https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_BUNDLE_INFO_PLIST.html

Screenshot at 2025-04-28 11-42-42


inflex:
Do not use this PR, see more recent fix which resolves it within the CMakeList directly

PR #339, opened 2025-10-17
Bump MacOS deplyment target to fix compilation error

k3a:
This fixes filesystem:path errors on recent MacOS: 

```
platform.h:26:19: error: 'path' is unavailable: introduced in macOS 10.15
   26 | const filesystem::path show_file_picker(bool filterBoards = false);
      |                   ^
```

ref: https://github.com/NixOS/nixpkgs/pull/452764

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.11)
 
 # Set OSX target version, before calling project() (inside version.cmake). FORCE is needed when project() is called within an included file
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X deployment version" FORCE)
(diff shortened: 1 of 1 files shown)


piernov:
`ghc::filesystem` is used as a fallback for older systems.
Check is performed in https://github.com/OpenBoardView/OpenBoardView/blob/11345118c0bd63142d02ac188aa5c9c2a15e165f/src/CMakeLists.txt#L155

The check for AppleClang is commented out to force the `ghc::filesystem` fallback when built on macOS. 
https://github.com/OpenBoardView/OpenBoardView/blob/master/src/openboardview/filesystem_impl.h is then used to select between `std::filesystem` or `ghc::filesystem` implementation.

The official macOS x86_64 release is built on Mac OS X 10.11 and runs fine there.

piernov:
IIRC this should have been fixed in 502d7ff.

k3a:
Pull request #340 replaces this

PR #343, opened 2025-12-14
Provide the ability to build a Flatpak package

### LINKED ISSUE #251: Please provide Flatpak / Flathub
https://docs.flatpak.org/en/latest/introduction.html
having a flat package ensures correct operation on any linux distribution, which makes it even easier to maintain.
Every day, more and more apps like flatpak are added, software like Kicad has been packaged as Flatpak for some time now.
This is also helpful for users of Fedora Silverblue and immutable systems.

vlad-sharpie:
I've made all the provisions for someone to build a Flatpak version of this great app by running `flatpak run --command=flathub-build org.flatpak.Builder --install --force-clean io.github.openboardview.openboardview.yml` in the root directory (if they've installed Flatpak Builder with `flatpak install -y flathub org.flatpak.Builder`). Alternatively for the user, the GitHub Actions workflow has been updated to build a package for them to download.

I tried to make everything as separated as possible from the current build process, so there is a separate folder with hardcoded details for app metadata instead of using the CMake variables (which I did because of differing naming conventions needed), but there is one check in `assets/CMakeLists.txt` that moves the corresponding files in `assets/flatpak` to the write place. 

This should fix #251, and the next step would be to integrate this with the Flathub repository! For the PDF integration to work I imagine we would need to do something with this: https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access


vlad-sharpie:
Here's it working, there's lots of errors relating to X11 but I can't test if that's actually a problem, it works well on Wayland.
image

piernov:
Hi, thanks for the contribution.

1. I do not think the asset files should be duplicated. CMake should configure and name them as needed in the target build instead.

2. We have used the `org.openboardview.openboardview` identifier in the past, I'd prefer sticking to that instead of `io.github.openboardview.openboardview`.

3. The `OBV_BUILD` depends on the git repository and needs to have the full history to count commits properly, which does not seem to happen here.

4. PDF bridge with Evince does not work (with the custom Evince Flatpak we provide), probably needs some additional configuration for D-Bus IPC to work.

5. During the build there is this warning which does not sound right (but maybe harmless):
```
CMake Warning:
  Manually-specified variables were not used by the project:

    FLATPAK
```

PerAstraAdDeum:
Hey

> I've made all the provisions for someone to build a Flatpak version of this great app by running `flatpak run --command=flathub-build org.flatpak.Builder --install --force-clean io.github.openboardview.openboardview.yml` in the root directory (if they've installed Flatpak Builder with `flatpak install -y flathub org.flatpak.Builder`). Alternatively for the user, the GitHub Actions workflow has been updated to build a package for them to download.
> 
> I tried to make everything as separated as possible from the current build process, so there is a separate folder with hardcoded details for app metadata instead of using the CMake variables (which I did because of differing naming conventions needed), but there is one check in `assets/CMakeLists.txt` that moves the corresponding files in `assets/flatpak` to the write place.
> 
> This should fix #251, and the next step would be to integrate this with the Flathub repository! For the PDF integration to work I imagine we would need to do something with this: https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access

Hey, thank you for your effort! I'd really appreciate if this would finally land on Flathub. I'm not a dev though, what's it stopping from building here?

Best regards

PR #346, opened 2026-01-26
Gencad files:  support for a 'testcad' variant

galkinvv:
Support a variant of gencad files with extra fields in header (including TESTCAD marker)
and spurious `COMPONENT` duplication for each pin (so for shapes with 1000 pins, component declaration is duplicated 1000 times)

Testing sample `gencad-sample-testcad` is available in [google drive SamplesForOBV folder](https://drive.google.com/drive/folders/1XD3OSYpvj5znjMmG58cHdOut3f2ltg--)


galkinvv:
This works, but still in draft because I'm not sure that all caching introduced is actually effective/useful (was added before the component-duplication issue was identified)

galkinvv:
To handle the test file with sane speed (3 seconds instead of 1 minute), the cache for fast padstack-by-name lookup was implemented.

However, making "load speed" for this sample file and gencad files in general much faster than 0.5&nbsp;Second/Megabyte seems to be blocked by the lack of an efficient way to parse numeric literals with the mpc library.

galkinvv:
Stambled on another file containing TEST in headers  and not openable due to extra strings in DEVICES.
`gencad-sample-test-in-header-and-extras-in-device.cad` added to same shared googledrive folder linked in the first message.

Fixed this in the same PR by allowing arbitrary strings in device section.

I'll use this version for day-to-day .cad viewing for some period and if no issues would be found - I will mark it ready for review

galkinvv:
Used this version for some time, no new bugs discovered so marking as "Ready to review"

piernov:
Let's merge as-is, make a release and see what catches fire.
Thanks for the contribution.

PR #349, opened 2026-03-05
Fix infinite loop caused by negative board surface size

### LINKED ISSUE #348: Setting DPI to 200 and an `infoPanelWidth` greater than twice the `windowX` in the config file causes the application to hang on startup on Linux
When `windowX=1200`, `dpi=200`, and  `infoPanelWidth = 652` in `obv.conf` on Linux, the application gets stuck in an infinite loop in `BoardView::OutlineGenFillDraw`. Bisecting shows the problem first appeared in e238d63a082aabcebd31cc0d9d5be864344d021e.

I already have a fix for this that I'll open a PR for, but I wanted to open an issue first since I'm not sure my solution is the best way to go about fixing this.

cyrozap:
After the change in e238d63, having `windowX=1200`, `dpi=200`, and `infoPanelWidth = 652` in `obv.conf` on Linux would cause `app.m_board_surface.x` to be negative, which caused `sx` to be negative, which caused `m_scale` to be negative, which caused `vdelta` to be negative, which caused `y` to never reach `yend`, causing an infinite loop. Fix this by detecting the negative condition early and setting `app.m_board_surface.x` and `app.m_info_surface.x` to reasonable values.

Fixes #348.

Fixes: e238d63 ("High-DPI support on Windows")

### DIFF
--- a/src/openboardview/main_opengl.cpp
+++ b/src/openboardview/main_opengl.cpp
@@ -352,6 +352,10 @@ int main(int argc, char **argv) {
 	app.m_board_surface.x = g.width;
 	app.m_board_surface.y = g.height;
 	if (app.config.showInfoPanel) app.m_board_surface.x -= app.m_info_surface.x;
+	if (app.m_board_surface.x <= 0.0f) {
+		app.m_board_surface.x = g.width * 0.66f;
+		app.m_info_surface.x = g.width - app.m_board_surface.x;
+	}
 
 	if (g.font_size > 0.0) app.config.fontSize = g.font_size;
 


piernov:
Thanks for the suggested fix. While it does take care of the problem, I'm not sure it is the correct way to take care of it.
That said, this sizing logic seem flawed to me anyway. Why does it rely on `g.width` which doesn't seem to be the actual drawable surface width (and then what about actual display DPI)? It sounds like there are many ways for things to go wrong here.
The root cause of an incorrect value in the config file probably comes from here though: https://github.com/OpenBoardView/OpenBoardView/blob/f17c1ecaec96a2a69da83b91a01108f92bbebc68/src/openboardview/BoardView.cpp#L248
What should be written to the config file is `config.infoPanelWidth`, value without the DPI scale, since the DPI scale is applied after loading back the value.
However I'm not sure if this actually prevents anything wrong from happening during execution, if something else could cause the loop bound to be wrong, then the loop itself should be protected in some way.

cyrozap:
Yeah, like I said in #348, I wasn't sure that this PR was the best way to fix the issue, so please feel free to implement a more "correct" fix and close this PR.

> The root cause of an incorrect value in the config file probably comes from here though:
> ...

I'm not sure I'd call an incorrect value in the config file the "root cause". IMO, bogus/incorrect/out-of-bounds values in the config file should never cause the application to hang or crash, as the config file could just as easily become corrupted by other means.

The reason the loop becomes infinite is because `app.m_board_surface.x` is negative. And the reason `app.m_board_surface.x` is becoming negative during execution is because `app.m_info_surface.x` is subtracted from it when `app.config.showInfoPanel` is set, so if `app.m_board_surface.x` is ever less than `app.m_info_surface.x` this subtraction will cause `app.m_board_surface.x` to become negative.

The incorrect configuration value write is definitely something that should also be fixed, but the main issue here IMO is that the code is not tolerant of incorrect values in the config file.

PR #351, opened 2026-03-07
FileFormats/{CAEFile,FZFile}: Correct component side placement

### LINKED ISSUE #350: Many parts are not being displayed for a particular FZ file
Many parts are not getting rendered for this file: [Asus X99-E WS USB 3.1 60SB04Q0-SB0A03.FZ](https://drive.google.com/file/d/1V3MDxNTtLRqLpzQeMYyhHSo2SQ5fRotf/view)

Screenshot of the problem:

Image

cyrozap:
When a component in an FZ or CAE file is mirrored, that means it's on the bottom of the board, not the top. Before this change, a board would have to be rotated 180 degrees, flipped, and then mirrored to be displayed in the correct orientation. After this change, FZ and CAE files are displayed correctly the moment they're opened, without requiring any user input.

Fixes #350

### DIFF
--- a/src/openboardview/FileFormats/FZFile.cpp
+++ b/src/openboardview/FileFormats/FZFile.cpp
@@ -359,9 +359,9 @@ void FZFile::parse(std::vector<char> &buf, const std::array<uint32_t, 44> &fzkey
 				/*char *srotate =*/READ_STR();
 				part.part_type = BRDPartType::SMD;
 				if (!strcmp(smirror, "YES"))
-					part.mounting_side = BRDPartMountingSide::Top; // SMD part on top
-				else
 					part.mounting_side = BRDPartMountingSide::Bottom; // SMD part on bottom
+				else
+					part.mounting_side = BRDPartMountingSide::Top; // SMD part on top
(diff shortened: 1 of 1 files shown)


galkinvv:
I can confirm that before this patch >80% of .fz files I had used in practice - forced me to mirror the board to get physically correct orientation. After this patch - they opens fine by default. I can't remember if there was any .fz files that opens fine before this, so maybe it was even 100% instead of 80%.

The need to flip is a minor problem (thats intuitive, most boards are often flipped during viewing), but the need to mirror - is major problem, since its not obvious: I know at least 2 different cases when people in the forum discussions were confused after reading the IC pinout in the mirrored boardview presentation.

Also, as far as i remeber I've seen the same need-to-mirror-bug in other software, maybe boardviewer (not sure).

I tested this patch with several boardviews - works great in practice - no more need for manual mirroring. So, manually tested & looks good to me.

Thanks, @cyrozap

piernov:
Sorry for the late answer,
@cyrozap thanks a lot for the fix
@galkinvv thanks for testing, I haven't had the time to look at it yet.

PR #354, opened 2026-05-02
annotations: defer SQLite file creation until first write

atillacv:
Previously, opening any board file would immediately create a `.sqlite3` file alongside it, even if the user never added any annotations. This clutters board directories for users who only view boards.

## Changes

- `Annotations::Load()`: use `sqlite3_open_v2` with read-write only (no create flag) if the file already exists; if it doesn't exist, keep `sqldb = nullptr` and return early
- `Annotations::Add()`: if `sqldb` is null, open/create the database file on demand before inserting the first annotation
- Added `sqlfilename` member to `Annotations` struct to store the path for deferred creation

## Result

No `.sqlite3` file is created when simply opening and viewing a board. The file is only created the first time the user adds an annotation, which is the expected behaviour.

### DIFF
--- a/src/openboardview/annotations.cpp
+++ b/src/openboardview/annotations.cpp
@@ -62,7 +62,14 @@ int Annotations::Load(void) {
 	if (pos != std::string::npos) sqlfn[pos] = '_';
 	sqlfn += ".sqlite3";
 
-	sqldb = nullptr;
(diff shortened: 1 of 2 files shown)


atillacv:
Closing to resubmit with improved implementation

PR #355, opened 2026-05-02
macOS: add Skim PDF bridge (PDFBridgeSkim)

atillacv:
Implements PDF bidirectional search on macOS using [Skim](https://skim-app.sourceforge.io/) as the PDF viewer, controlled via AppleScript through `osascript`.

## Features

- **Auto-open PDF**: when a BRD file is loaded, the matching PDF is opened automatically in Skim. Lookup order: same directory as BRD → configurable `pdfDirectory` fallback
- **OBV → Skim search**: `DocumentSearch()` finds a component/net in Skim and navigates to it, cycling through multiple occurrences across pages using `find doc text … from character 1 of page N`
- **Skim → OBV search**: `HasNewSelection()` polls Skim every 2 seconds (non-blocking) for the current text selection
- **PDF folder / BRD folder settings** added to Program Preferences (previously these fields only appeared on Windows — moved outside `#ifdef _WIN32`)
- `show_folder_picker()` implemented on macOS via `NSOpenPanel` in `osx.mm`
- `PDFFile::loadFromConfig()` accepts an optional `pdfSearchDir` fallback parameter

## Implementation notes

- Uses `system("osascript /tmp/obv_skim_script.applescript")` instead of `NSAppleScript` to avoid activation/find issues
- `PDFBridgeSkim.mm` compiled with `-fobjc-arc`
- Skim must be installed (`brew install --cask skim`)

## Testing

Tested on macOS (Apple Silicon) with BRD files from `/Volumes/…/boardview/` and matching PDFs in a separate directory configured via Preferences.

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -158,7 +158,7 @@ int BoardView::LoadFile(const filesystem::path &filepath) {
 				auto conffilepath = filepath;
 				conffilepath.replace_extension("conf");
 				backgroundImage.loadFromConfig(conffilepath);
-				pdfFile.loadFromConfig(conffilepath);
+				pdfFile.loadFromConfig(conffilepath, filesystem::u8path(config.pdfDirectory));
 
 				pdfBridge.OpenDocument(pdfFile);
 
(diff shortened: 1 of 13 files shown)


atillacv:
Closing to resubmit with improved implementation

PR #356, opened 2026-05-04
annotations: defer SQLite file creation until first write

atillacv:
## Summary
- Do not create `.sqlite3` annotation file when a BRD is opened
- File is only created when the user actually adds an annotation
- Prevents cluttering the boardview directory with empty database files

## Test plan
- [ ] Open a BRD file — no `.sqlite3` file should appear
- [ ] Add an annotation — `.sqlite3` file is created at that point

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/openboardview/annotations.cpp
+++ b/src/openboardview/annotations.cpp
@@ -62,7 +62,14 @@ int Annotations::Load(void) {
 	if (pos != std::string::npos) sqlfn[pos] = '_';
 	sqlfn += ".sqlite3";
 
-	sqldb = nullptr;
+	sqldb      = nullptr;
+	sqlfilename = sqlfn; // save for deferred creation in Add()
+
+	// Only open if the file already exists — don't create it on load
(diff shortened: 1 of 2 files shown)


galkinvv:
(non-mantainer comment)

Looks useful feature, as a day-to-day user those unused .sqlite file are quite annoying. Thanks for suggesting!

This looks quite compact and reviwebale, however there are some nitpicks to make it even more simple/mantainable:

- separate sqlfilename member as addition to local variable looks supeficial, maybe just uplift existing local variable to member
- the file existance check can be made much more readable like it is done in PDFBridge folder - including `<filesystem_impl.h>` and using `filesystem::exists`
- the sqldb opening code duplication can be avoided by moving it inside Init()

P.S. I like this PR, although I'm the same one who criticized the adjacent PR as unreviewable.
So I'm not against AI as an approach.

PR #357, opened 2026-05-04
macOS: Skim PDF bridge with bidirectional sync

atillacv:
## Summary
- Adds `PDFBridgeSkim` — a macOS-specific PDF bridge using Skim via AppleScript
- **Skim → OBV**: selecting text in Skim automatically searches and highlights the component in OBV (~500ms latency, background thread, no UI blocking)
- **OBV → Skim**: clicking PDF Search in OBV opens the matching schematic in Skim and navigates to the component
- **Auto-close**: Skim documents opened by OBV are closed when OBV exits
- Fix `exit()` → `return` in main loop so C++ destructors run on quit

## Technical details
- Background polling thread reads Skim's text selection every 500ms via AppleScript
- Also monitors `/tmp/obv_search_request.txt` for integration with macOS Services (right-click menu)
- Fixed AppleScript selection extraction: use `repeat with s in sel` instead of `string of sel`
- Search result written to `m_search[0]` so highlights persist across render frames
- Thread is detached on exit; Skim close script runs synchronously before process ends

## Test plan
- [ ] Open a BRD in OBV — matching PDF opens in Skim automatically
- [ ] Select a component name in Skim — OBV highlights it within ~1 second
- [ ] Click PDF Search in OBV — Skim navigates to the component
- [ ] Close OBV — Skim documents opened by OBV close automatically

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -158,7 +158,7 @@ int BoardView::LoadFile(const filesystem::path &filepath) {
 				auto conffilepath = filepath;
 				conffilepath.replace_extension("conf");
 				backgroundImage.loadFromConfig(conffilepath);
-				pdfFile.loadFromConfig(conffilepath);
+				pdfFile.loadFromConfig(conffilepath, filesystem::u8path(config.pdfDirectory));
 
 				pdfBridge.OpenDocument(pdfFile);
 
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -3573,7 +3573,11 @@ void BoardView::HandlePDFBridgeSelection() {
 		if (selection.empty()) {
 			m_pinHighlighted.clear();
 			m_partHighlighted.clear();
+			m_search[0][0] = '\0';
 		} else {
+			// Update search box so highlights persist across frames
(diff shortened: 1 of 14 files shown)


atillacv:
Note: The Windows build failures (crossbuild-32bit-exe-on-mingw, build-64bit-exe-on-windows) are pre-existing issues unrelated to this PR — they fail on the same `show_folder_picker` linker error that exists in master. All macOS and Linux builds pass.

galkinvv:
(non-mantainer comment)

> are pre-existing issues unrelated to this PR — they fail on the same `show_folder_picker` linker error that exists in master.

No. Just No. If you want other people to pay attention to AI-assisted code you suggest - be careful to not loose the trust - dont write PR comments that need to be rechecked.

To the technical side of PR - instruct AI make small & reviewable changes with minimal needed changes to existing classes and using english comments and features like folder selection split to different PRs. It can't be reviewed in current form - many changes just looks spurious, like brdDirectory addition, making overall project even harder to mantain.

Even the "exit() → return" change is questinable. Openboardview is a simple gui app, not a enterprise producation server, and making a fast exit is in general safer for hangs/crashes then running all unneeded destructors. If you need to close something before exit - just make explicit call instead of relying on destructor.

PR #363, opened 2026-07-29
Add a macOS PDF bridge backend (Skim)

### LINKED ISSUE #285: pdf  search
hello im having issue using pdf search button, when i click it nothiing happens, pdf already opened, im using mac.

IamMrCupp:
Fixes #285.

## What this does

Today the PDF search button does nothing on macOS: the `#ifdef` chain in `BoardView.h` has no macOS branch, so it falls through to the base `PDFBridge` — whose methods are all empty no-ops. The click calls into nothing and returns. It's an unimplemented backend, not a bug.

This adds `PDFBridgeSkim`, a backend that drives [Skim](https://skim-app.sourceforge.io/) over AppleScript — the same "drive an existing viewer" model as the Evince (Linux) and SumatraPDF (Windows) backends. Both directions of cross-probing work:

- **Forward** (board → schematic): clicking a part's *PDF Search* jumps Skim to that designator and highlights it.
- **Reverse** (schematic → board): selecting text in Skim highlights the matching part on the board.

## Why Skim

It's open-source, actively maintained, and scriptable out of the box — `find`, `selection`, `go`, `obtain` cover everything the bridge needs, so there's nothing to patch. That's the same reasoning behind picking Evince and SumatraPDF for the other platforms.

## How it works

- **Forward search / open** run synchronously via `NSAppleScript` on the render thread — they only fire on a button press, so a brief round-trip is fine.
- **The document is resolved by path and opened on demand** on every operation, so a search still works if the user has closed the Skim window or has other PDFs open (no reliance on `document 1`).
- **Reverse search polls** Skim's selection on a background thread — Skim has no cross-process selection-change signal, so polling is the option. It runs off the render loop and writes to a mutex-guarded value that `HasNewSelection()` only compares, never blocking. Interval is configurable via `pdfSkimReverseSearchPollMs` (default 300 ms; `0` disables reverse search). The poll uses an `osascript` subprocess because `NSAppleScript` is main-thread affine.
- **The poll wakes the render loop** with `SDL_PushEvent` when the selection changes — OBV skips `Update()` after it's been idle a while, so without the nudge a background selection wouldn't be picked up until the next input event.
- **`NSAppleEventsUsageDescription`** is added to the bundle Info.plist — without it the automation events fail silently (which reproduces the exact "button does nothing" symptom).
- Skim's `find` has no whole-words option, so `wholeWordsOnly` searches as a substring — logged at debug rather than silently ignored.

Everything macOS-specific is guarded by `ENABLE_PDFBRIDGE_SKIM` (defined in the `if(APPLE)` CMake branch) or `if(APPLE)`, so the Linux and Windows builds are untouched. `PDFBridgeSkim.mm` is compiled with ARC; the rest of the ObjC++ stays manual retain/release.

## Commits

1. `fix(PDFBridge): warn when no PDF bridge backend is available on this platform` — surfaces the silent failure on any platform without a backend, instead of the click doing nothing.
2. `feat(PDFBridge): add macOS Skim backend for boardview-schematic cross-probing` — the backend, CMake wiring, `BoardView.h` dispatch branch, and the Info.plist entitlement.

## Testing

Built and tested on an M4 Pro, macOS 26.5.2, Skim 1.7.15, against a real board + matching schematic PDF. Forward search jumps Skim to the clicked part; reverse search highlights the board part from a Skim text selection, hands-free. Built arm64; I don't have an Intel Mac to verify the x86_64 half of a universal build, but nothing here is arch-specific.

Happy to adjust the approach — polling cadence, config keys, whatever — if you'd rather it work differently.

### DIFF
--- a/asset/MacOSXBundleInfo.plist.in
+++ b/asset/MacOSXBundleInfo.plist.in
@@ -32,5 +32,7 @@
 	<string>${MACOSX_BUNDLE_COPYRIGHT}</string>
 	<key>NSHighResolutionCapable</key>
 	<true/>
+	<key>NSAppleEventsUsageDescription</key>
+	<string>OpenBoardView controls a PDF viewer (Skim) to search the schematic for the selected component.</string>
 </dict>
 </plist>
--- a/src/openboardview/BoardView.h
+++ b/src/openboardview/BoardView.h
@@ -20,6 +20,7 @@
 #include "PDFBridge/PDFBridge.h"
 #include "PDFBridge/PDFBridgeEvince.h"
 #include "PDFBridge/PDFBridgeSumatra.h"
+#include "PDFBridge/PDFBridgeSkim.h"
 #include "PDFBridge/PDFFile.h"
 #include <cstdint>
 #include <vector>
@@ -64,6 +65,8 @@ struct BoardView {
 	PDFBridgeEvince pdfBridge;
 #elif defined(_WIN32)
 	PDFBridgeSumatra &pdfBridge = PDFBridgeSumatra::GetInstance(obvconfig);
+#elif defined(ENABLE_PDFBRIDGE_SKIM)
+	PDFBridgeSkim pdfBridge{obvconfig};
 #else
 	PDFBridge pdfBridge; // Dummy implementation
 #endif
--- a/src/openboardview/CMakeLists.txt
+++ b/src/openboardview/CMakeLists.txt
@@ -17,6 +17,7 @@ else()
 		find_library(COCOA_LIBRARY Cocoa)
 		find_package(OpenGL REQUIRED)
 		set(OPENGL_INCLUDE_DIR ${OPENGL_INCLUDE_DIR}/Headers)
+		add_definitions(-DENABLE_PDFBRIDGE_SKIM) # macOS PDF bridge via Skim
 	else(APPLE)
 		find_package(PkgConfig REQUIRED)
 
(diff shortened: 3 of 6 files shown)


IamMrCupp:
Heads-up for anyone building this on macOS to try it — there's a crash you can hit that isn't from this patch.

Homebrew has migrated `sdl2` to **sdl2-compat** (2.32.70 currently), which reimplements the SDL2 API on top of SDL3 and `dlopen`s `libSDL3.dylib` from its own initializer. `fixup_bundle` copies *linked* dependencies only, so a `dlopen`'d library never lands in the bundle. The `.app` then aborts before `main`:

```
SIGABRT  libSDL2-2.0.0.dylib `dllinit`
         <- dyld4::Loader::findAndRunAllInitializers
asi: abort() called
```

The shim's own message is `Failed loading SDL3 library.`

Workaround — drop SDL3 in next to the shim, whose first search path is `@loader_path/libSDL3.dylib`, then re-sign since the copy invalidates the signature:

```sh
APP=release_build/src/openboardview/openboardview.app
cp -L "$(brew --prefix)/lib/libSDL3.0.dylib" "$APP/Contents/Frameworks/libSDL3.dylib"
chmod u+w "$APP/Contents/Frameworks/libSDL3.dylib"
install_name_tool -id "@loader_path/libSDL3.dylib" "$APP/Contents/Frameworks/libSDL3.dylib"
codesign --force --deep --sign - "$APP"
```

One file is enough — SDL3 3.4.12 pulls in nothing outside the system frameworks.

This is orthogonal to the Skim bridge; it hits any current-Homebrew macOS build of `master`. CI isn't affected either — the universal job uses the libsdl.org SDL2 framework and the Intel job uses the `brew_oldermacos` tap, so released DMGs are fine. Flagging it purely so it doesn't get read as a fault in this PR.

IamMrCupp:
actively using this on a few different M series Macs without any issues. I'll test the build on an Intel Mac once I pull my older mac out of the storage bin and get it online for testing.

PR #366, opened 2026-08-10
Add support for Tebo-ICT View (.tvw) board files

### LINKED ISSUE #291: Teboview (tvw) format support request
Noticed that there isn't an issue about it yet in this repo.

* https://www.eevblog.com/forum/eda/wanted-teboview-tvw-file-format-specification-to-write-openboardview-decoder/
* https://github.com/inflex/teboviewformat

MARCEBELE:
Fixes #291.

## What this does

OpenBoardView currently rejects Tebo-ICT View (`.tvw`) files with
"Unrecognized file format.". This adds a native reader 

Copper/plane rendering is **not** part of this PR, it's a separate,
larger change on top of this one.

## Where the parser comes from

The binary container parser in `FileFormats/tebo/` is Pavel Kovalenko's
reverse-engineered implementation from
[eagleview](https://github.com/nitrocaster/eagleview), headers reformatted to match this project's
`Crypto/des.c` convention). `FileFormats/TVWFile.cpp` is a thin adapter
mapping the parsed board onto `BRDFileBase`.

Changes relative to upstream eagleview:
- assertions throw instead of calling `__builtin_trap()`, so a malformed
  file surfaces as a load error instead of taking the application down
- an empty through-layer (`toolCount == 0`) is accepted; upstream
  asserted, which rejected most of the Lenovo/LCFC boards tested
- diagnostics are behind `TEBO_VERBOSE` (off) rather than `printf`
- kept C++11-compatible, since this target builds as C++11

## How it works

- **Detection** is by content, not extension: every `.tvw` begins with a
  length-prefixed, character-ciphered type string that decodes to
  `"Tebo-ictview files."`.
- **Pin → net resolution** is two hops, undocumented in the file itself:
  `Pin.Handle / 8` indexes into that layer's `Pads` array, and
  `Pad.Net` indexes into the board's net table. A `Pin` record carries
  no net field directly.
- **Per-layer field order varies.** A layer's header carries a flag
  (`1` = normal order, `2` = extra data appended after surfaces) that
  changes where lines/arcs are re-read from. Getting this wrong desyncs
  the stream silently for every layer after it, so it's asserted rather
  than assumed.
- **Board outline** comes from the real route/fab layer (`Roul`/OUTLINE)
  when present; falls back to a bounding box around all pins when it
  isn't, or when the route data looks unusable (stray far-out segments).
- Net names are interned once into a `std::deque<std::string>` (not
  `std::vector`) so pointers handed out via `intern()` survive later
  growth.

## Testing

9 boards (Lenovo Yoga/Legion/ThinkBook, LCFC), 22k pins on the largest —
all load, every pin resolves to a named net. Malformed/truncated inputs
were fuzzed and produce a clean load error rather than a crash.

### DIFF
--- a/src/openboardview/BoardView.cpp
+++ b/src/openboardview/BoardView.cpp
@@ -28,6 +28,7 @@
 #include "FileFormats/CSTFile.h"
 #include "FileFormats/FZFile.h"
 #include "FileFormats/GenCADFile.h"
+#include "FileFormats/TVWFile.h"
 #include "FileFormats/XZZPCBFile.h"
 #include "GUI/DPI.h"
 #include "GUI/Fonts.h"
@@ -140,6 +141,8 @@ int BoardView::LoadFile(const filesystem::path &filepath) {
 				m_file = new BRDAllegroFile(buffer);
 			else if (XZZPCBFile::verifyFormat(buffer))
 				m_file = new XZZPCBFile(buffer, config.XZZPCBKey);
+			else if (TVWFile::verifyFormat(buffer))
+				m_file = new TVWFile(buffer);
 			else
 				m_error_msg = "Unrecognized file format.";
 
--- a/src/openboardview/CMakeLists.txt
+++ b/src/openboardview/CMakeLists.txt
@@ -82,6 +82,8 @@ set(SOURCES
 	FileFormats/CSTFile.cpp
 	FileFormats/FZFile.cpp
 	FileFormats/GenCADFile.cpp
+	FileFormats/TVWFile.cpp
+	FileFormats/tebo/TeboBoard.cpp
 	FileFormats/XZZPCBFile.cpp
 	NetList.cpp
 	PartList.cpp
--- a/src/openboardview/FileFormats/TVWFile.cpp
+++ b/src/openboardview/FileFormats/TVWFile.cpp
@@ -1,0 +1,167 @@
+#include "TVWFile.h"
+
+#include "tebo/TeboBoard.hpp"
+
+#include <algorithm>
+#include <cstdint>
+#include <sstream>
+
+// Every .tvw begins with a length-prefixed, char-ciphered type string that
+// decodes to "Tebo-ictview files." — the raw bytes are a stable signature.
+bool TVWFile::verifyFormat(std::vector<char> &buf) {
+	static const char sig[] = {0x13, 'O', '9', '5', 'w', '-', '2', '8', 'p', 's',
+	                           '4', '9', 'm', ' ', '0', '2', 'v', '9', 'o', '.'};
+	if (buf.size() < sizeof(sig)) return false;
+	return std::equal(std::begin(sig), std::end(sig), buf.begin());
+}
+
+const char *TVWFile::intern(const std::string &s) {
+	pool_.push_back(s);
+	return pool_.back().c_str();
+}
+
+namespace {
+// Tebo coordinates are 2-decimal fixed point; convert to integer mil like the
+// reference Toptest exporter (Fixed32 -> int via float, i.e. raw/100).
+inline BRDPoint ToPoint(Tebo::Vector2S v) {
+	return BRDPoint(static_cast<int>(static_cast<float>(v.X)),
+	                static_cast<int>(static_cast<float>(v.Y)));
+}
+} // namespace
+
+TVWFile::TVWFile(std::vector<char> &buf) {
+	try {
+		std::string data(buf.begin(), buf.end());
+		std::istringstream is(data, std::ios::in | std::ios::binary);
+
+		Tebo::Board board;
+		board.Read(is);
+
+		// Locate the top and bottom copper (logic) layers; parts/pins live there.
+		int topIdx = -1, botIdx = -1;
+		for (size_t i = 0; i < board.Layers.size(); i++) {
+			auto *lo = dynamic_cast<Tebo::LogicLayer *>(board.Layers[i].get());
+			if (!lo) continue;
+			if (lo->Type == Tebo::LayerType::Top) topIdx = static_cast<int>(i);
+			else if (lo->Type == Tebo::LayerType::Bottom) botIdx = static_cast<int>(i);
+		}
+
+		// Real board outline: the route ("Roul"/OUTLINE) layer stores the board
+		// edge as milled slots. Collect them now, sanity-filter against the pin
+		// extents once those are known (some files contain stray segments), and
+		// fall back to a bounding box if nothing usable survives.
+		std::vector<std::pair<BRDPoint, BRDPoint>> rawOutline;
+		for (auto const &lp : board.Layers) {
+			auto *tl = dynamic_cast<Tebo::ThroughLayer *>(lp.get());
+			if (!tl || tl->Type != Tebo::LayerType::Roul) continue;
+			for (auto const &slot : tl->DrillSlots)
+				rawOutline.emplace_back(ToPoint(slot.Begin), ToPoint(slot.End));
+		}
+
+		// Intern net names once so pins can share the pointers.
+		std::vector<const char *> netPtr(board.Nets.size(), nullptr);
+		for (size_t i = 0; i < board.Nets.size(); i++)
+			netPtr[i] = intern(board.Nets[i]);
+
+		auto layerOf = [&](uint32_t partLayer) -> Tebo::LogicLayer * {
+			int idx = -1;
+			if (static_cast<int>(partLayer) == topIdx) idx = topIdx;
+			else if (static_cast<int>(partLayer) == botIdx) idx = botIdx;
+			if (idx < 0) return nullptr;
+			return dynamic_cast<Tebo::LogicLayer *>(board.Layers[idx].get());
+		};
+
+		BRDPoint bmin{INT32_MAX, INT32_MAX}, bmax{INT32_MIN, INT32_MIN};
+
+		for (Tebo::Part const &part : board.Parts) {
+			Tebo::LogicLayer *layer = layerOf(part.Layer);
+			if (!layer) continue; // not on top/bottom copper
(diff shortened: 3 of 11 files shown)


rzsn:
I have tested the fork, and it works for me too!
Have not found anything unusual when I was working on a laptop repair task.
Would appreciate the inclusion of this PR.

Test file was `MY7A3 MY7A4 MY760 MY56C NM-G611.tvw` - an LCFC made _Lenovo Legion Pro 7 16IAX10H_ unit.

piernov:
Thanks for the contribution, unfortunately I do not have the time to review it properly for now.

Few questions in the meantime though:
1. Why the choice of not implementing just the needed bits of the parser directly in OBV like other formats? There are a bunch of features not supported into OBV as of now so I am not sure we need to implement decoding of every bits and pieces.
2. If it makes more sense to reuse the existing code from [nitrocaster/eagleview](https://github.com/nitrocaster/eagleview), would it be possible to use it as a submodule instead of copying the files into OBV?
3. Related to the previous point, have you tried to contact the original author to have better interoperability? For example it could be nice to profit from their parser for Eagle as well, so maybe use it as a library and just have a small shim in OBV to call into it for both formats.

Raihan8189:
1. Why not implement only the needed bits directly?

the container. `.tvw` is a forward-only .NET `BinaryReader` stream: the parser operates sequentially, it reads every byte in order, and the header carries only three sizes and a layer count and there no offset table. So features you don't care about still has to be decoded to find where the next one starts. Skipping it it desyncs everything after it.

There is also a per-layer flag (1 = normal ordering, 2 = extra data appended after the surfaces) that controls where lines and arcs are re-read from.

- Use in this PR: Layers, nets, parts, pins, pads, drill slots, pad shapes/positions, and board outline.

- set-up for copper features: Shapes, lines, arcs, surfaces (plane pours and voids), tools, and drill holes—essentially all the remaining geometry.

Most of what looks like dead weight here is required setup for the copper parser, which I kept here rather than the next PR.
Screenshot 2026-08-31 005511
Screenshot 2026-08-31 010521

2. Submodule instead of copying the files?

it doesn't work asis:
- eagleview is a standalone CLI application, not a library. There's no parser target to link against, so it would mean submoduling a whole application to use seven files
- It's C++17 The not plug and play compatible with OBV 11
- There are behavioural changes too that i made so it would not crash the whole app
Assertions now throw exceptions instead of calling __builtin_trap(), ensuring a malformed file surfaces as a load error rather than crashing the application and empty through layers are now accepted where upstream would assert change that make  Lenovo/LCFC boards I tested working.

> Thanks for the contribution, unfortunately I do not have the time to review it properly for now.
> 
> Few questions in the meantime though:
> 
> 1. Why the choice of not implementing just the needed bits of the parser directly in OBV like other formats? There are a bunch of features not supported into OBV as of now so I am not sure we need to implement decoding of every bits and pieces.
> 2. If it makes more sense to reuse the existing code from [nitrocaster/eagleview](https://github.com/nitrocaster/eagleview), would it be possible to use it as a submodule instead of copying the files into OBV?
> 3. Related to the previous point, have you tried to contact the original author to have better interoperability? For example it could be nice to profit from their parser for Eagle as well, so maybe use it as a library and just have a small shim in OBV to call into it for both formats.

MARCEBELE:
1. Why not implement only the needed bits directly?

the container. `.tvw` is a forward-only .NET `BinaryReader` stream: the parser operates sequentially, it reads every byte in order, and the header carries only three sizes and a layer count and there no offset table. So features you don't care about still has to be decoded to find where the next one starts. Skipping it it desyncs everything after it.

There's also a per-layer flag (1 = normal ordering, 2 = extra data appended after the surfaces) that changes where lines and arcs are re-read from.

- used here: layers, nets, parts, pins, pads, drill slots, pad shapes and
  positions, board outline
- used by the follow-up copper : shapes, lines, arcs, surfaces (plane pours
  and their voids), tools, drill holes i.e. all the geometry
So most of what looks like dead weight here is the copper feature's input i put them here rather the doing them in next pr
Screenshot 2026-08-31 010521
Screenshot 2026-08-31 010453

2. Submodule instead of copying the files?

it doesn't work asis:
- eagleview is a standalone CLI application, not a library. There's no parser
  target to link against, so it would mean submoduling a whole application to
  use seven files
- It's C++17 not plug and play compatible with OBV 11
- There are behavioural changes too that i made so it would not crash the whole app
Assertions now throw exceptions instead of calling __builtin_trap(), ensuring a malformed file surfaces as a load error rather than crashing the application. Additionally, empty through layers are now accepted where upstream would assert change that alone compatibility for most of the Lenovo/LCFC boards I tested.

IamMrCupp:
Built and tested this on macOS arm64 (M4 Pro, master `5e8d3b6` + `de96645`) — builds clean, no warnings from the new code, content-based detection slots in fine, and the MIT headers from eagleview are intact. Then I pulled six real `.tvw` boards off badcaps to see how it holds up outside the nine you tested. Two findings, one of which matters a lot.

## 1. Header layout: two of three Lenovo boards fail to load

Only one of the six loads as-is:

| Board | Customer (decoded) | Result on PR |
|---|---|---|
| LCFC GY535 NM-D191 Rev0.2 | LianBao | **loads** — 3948 parts / 11569 pins |
| ThinkPad T490S NM-B891 | LianBao | `Unrecognized ObjectType` at TeboBoard.cpp:640 |
| Legion Y700 NM-C371 R03 | LianBao | `Unrecognized ObjectType` at TeboBoard.cpp:640 |
| Gigabyte GA-H110M-S2 Rev1.0 | Landrex | `dummy[0] == 0` at TeboBoard.cpp:984 |
| Gigabyte GTX 1080 V1.0 | Landrex | same |
| Gigabyte GV-N2070GAMING-8GC | Landrex | same |

Cause is in `TvwHeader::Load`. eagleview models the bytes after `Customer` as `Const2` (u8, always 0), `Date`, then `Const3` (3 × u8, always 0). Hex of the real files says otherwise — those are two more length-prefixed, ciphered strings around a u16:

```
LCFC     (loads):  ... 07 "G34vS4z"  00           0e <date>  00 00  00           16 00 00 00 ...
T490S    (fails):  ... 07 "G34vS4z"  03 "YZM"     11 <date>  00 00  06 "YZMded"  16 00 00 00 ...
Gigabyte (fails):  ... 07 "G5u9k8s"  08 "B!Z@6sob" 10 <date>  00 00  00           15 00 00 00 ...
```

So the real layout is `Customer, StrA, Date, u16, StrB, Size1, Size2, Size3, LayerCount`. When both extra strings are empty (every board you tested, and my LCFC one) that collapses to exactly `00` + `00 00 00`, which is why it looked constant. The moment either string is populated the sizes and layer count are read from the wrong offset and the layer loop starts on garbage.

Fix, in `TeboBoard.hpp`/`.cpp`:

```cpp
// TvwHeader
std::string Extra1;   // was: uint8_t Const2
std::string Date;
uint16_t    Const2;   // was: uint8_t Const3[3]
std::string Extra2;

// TvwHeader::Load
Customer = r.ReadString255(); DecodeString(Customer);
Extra1   = r.ReadString255(); DecodeString(Extra1);
Date     = r.ReadString255(); DecodeString(Date);
Const2   = r.ReadU16();
Extra2   = r.ReadString255(); DecodeString(Extra2);
```

With that, all three Lenovo boards load: T490S 3402 parts / 10094 pins, Legion 4717 / 14053, LCFC unchanged. Decoded, the extra strings are `DFT` / `DFT002` on the Lenovo files and `G!G@byt3` on Gigabyte's — vendor/tool tags, nothing structural.

The Gigabyte files get past the header with that fix (all 14–20 layers and 1728 nets parse) but then throw inside `MysteriousBlock::Load`. Their `Size1` is 21 where every Lenovo file is 22, so I'd guess that's a format version and the post-netlist section differs. I don't think it needs to block this PR — Lenovo/LCFC is the stated target — but it'd be worth having the error mention the unsupported variant rather than a bare assert, since Gigabyte desktop/GPU boards are most of what turns up when you search for `.tvw`.

## 2. Short reads go unchecked, so a truncated file reads uninitialized memory

`StreamReader` never looks at `gcount()` after `is.read()`. Past EOF the destination is left untouched and every later value — counts, string lengths, indexes — is whatever was on the stack. Mostly an `R_ASSERT` trips by luck, but a garbage *count* is just a loop bound.

Repro — 46 bytes, valid signature, zero layers, huge net count:

```python
import struct
sig = bytes([0x13]) + b'O95w-28ps49m 02v9o.'
hdr = sig + struct.pack('<I',0) + b'\x00\x00\x00\x00\x00\x00' + struct.pack('<IIII',0,0,0,0)
open('bad.tvw','wb').write(hdr + b'\x00'*16 + struct.pack('<II',0xFFFFFFF0,0xFFFFFFF0))
```

`ReadNetList` reserves ~4G strings and loops `ReadString255()` on a dead stream. Under ASan that hit 8.4 GB RSS in 60 s and was still climbing; in the real app it's a hang → OOM rather than a "TVW parse error". A corrupted download does the same thing by accident.

Fix is two asserts in `tebo/StreamReader.hpp`:

```cpp
uint8_t ReadU8()
{
    uint8_t r;
    is.read((char *)&r, 1);
    R_ASSERT(is.gcount() == 1);
    ...
}

template <typename T>
void Read(T *dst, size_t count)
{
    std::streamsize const itemSize = sizeof(T);
    is.read((char *)dst, itemSize*count);
    R_ASSERT(is.gcount() == itemSize * static_cast<std::streamsize>(count));
    ...
}
```

With that the repro fails in under 2 s with a clean parse error, and 300 random-tail mutations of a valid header all fail cleanly too.

---

Everything else reads fine — throw-instead-of-trap is the right call, the `std::deque` interning is correct, and the outline sanity filter makes sense (none of my six files had a `Roul` layer, so they all hit the bbox fallback). Happy to push both fixes as a PR against your branch if that's easier than patching by hand.

PR #372, opened 2026-09-06
Remove glad dependency.

apteryks:
Since ImGui commit e3988a84db ("Removed support for gl3w, glew, glad,
glad2, glbinding2, glbinding3 (all now unnecessary)"), the
backends/imgui_impl_opengl3.h now implements its own loader, which
makes glad redundant for all but the legacy OpenGL 1.1 loader. This
change removes glad along the legacy OpenGL 1.1 loader.


apteryks:
CI as good as for the last commit in the repo.

piernov:
There is no plan to remove the OpenGL 1 renderer in the foreseeable future, so unfortunately these changes will not be merged. If we had a good enough software fallback (which is one use case of OpenGL 1 on Windows when only the native software rendering is available) we *might* consider removing OpenGL 1.
Thanks for contribution anyway.

apteryks:
Hm, I had assumed it was legacy; out of curiosity, on which system(s) is OpenGL 1 still needed/used?

piernov:
For the OpenGL 3 renderer, the minimum required version is OpenGL 3.2.

I've had to workaround a font issue due to limited texture size for someone with a Radeon X550 (yes without the R, the one from 2005) on Windows in 2022, this one might be retired by now but there are still people running some Intel HD Graphics from the late 2000s/early 2010s which may not support OpenGL 3.2 properly (especially on Windows and with outdated drivers).
Heck you might even have people try to run on server/workstation integrated graphics without 3D acceleration.

That's on the hardware side, but on the software side you have people with outright broken drivers, running in virtual machines without GPU access, using remote desktop, etc…

Nowadays on Linux the software rendering is pretty good (you get OpenGL 4.6 through LLVMpipe) in case you cannot get hardware-accelerated rendering, but as far as I know it's not the case on Windows and you can only rely on OpenGL 1.1 provided by the built-in "GDI Generic" driver.

In any case, if for whatever reason an OpenGL 3.2 context cannot be created, trying to create an OpenGL 1.1 context is a good fallback as OBV can still be used properly with OpenGL 1.1.

All this is just to have the best chances of it working out of the box.

apteryks:
Thanks for the explanation! An idea of a change I could make to this PR would be to conditionally avoid using glad/building the opengl 1 backend, only when a user has explicitly configured (CMake) `-DENABLE_GL1=OFF`; this would make it a bit more minimal/clean on GNU/Linux distributions, and the path toward eventually dropping glad/GL1 would already be implemented.

